PR #1, opened 2023-09-05
CI: Add build & test Fedora workflow for GitHub Actions

vmihalko:


vmihalko:
Thank you for your comment and advice. I will proceed with the suggested changes in a separate pull request.

PR #415, opened 2024-01-03
ci: run build & unit tests in GH Actions

mrc0mmand:
Introduce a GH Actions job that builds & tests polkit in three
"categories":

1) Build with various optimization levels & other build-time options
   (both clang & gcc)

2) Build & run the unit tests (both clang & gcc)

3) Build & run the unit tests with Address Sanitizer and
   Undefined Behavior Sanitizer (only gcc for now)

There's a couple of outstanding issues that should be taken care of in
the near future:

 - get rid of the gazillion of warnings during compilation so we can
   build with -Werror (--werror)
 - fix build with clang >= 17
 - make test-polkitbackendjsauthority work under ASan

### DIFF
--- a/.github/workflows/ci.sh
+++ b/.github/workflows/ci.sh
@@ -1,0 +1,86 @@
+#!/bin/bash
+set -eux
+set -o pipefail
+
+# TODO
+#   - enable --werror
+#       - currently there's a lot of warnings which need to be taken care of first
+#   - re-enable test-polkitbackendjsauthority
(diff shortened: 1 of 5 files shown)


mrc0mmand:
An example of such CI run can be seen in https://github.com/mrc0mmand/polkit/pull/1.

vmihalko:
> fix build with clang >= 17

Soon, we plan to remove/replace `mocklibc`.

vmihalko:
Thank you for this PR!

PR #417, opened 2024-01-18
meson: decouple session tracking mechanism from unit and sysusers

bluca:
Session tracking build options are about whether logind or other APIs
are used in the polkit code, at compilation and linking time.

units, sysusers and tmpfiles files have nothing to do with code changes
or APIs, they simply install config files, that can just be ignored if
they are not needed. Use pkg-config if available, but otherwise have
a hard-coded fallback with the well-known defaults.

This also fixes another bug,  'systemdsystemunitdir' is specified as
an option the systemd_dep variable is not defined, but the sysusers.d
directory lookup uses it, causing a build failure:

```
dh_auto_configure -- \
	-Dexamples=false \
	-Dintrospection=true \
	-Dman=true \
	-Dsystemdsystemunitdir=/usr/lib/systemd/system \
	-Dtests=true \
	-Dgtk_doc=true -Dsession_tracking=libsystemd-login
	cd obj-x86_64-linux-gnu && DEB_PYTHON_INSTALL_LAYOUT=deb LC_ALL=C.UTF-8 meson setup .. --wrap-mode=nodownload --buildtype=plain --prefix=/usr --sysconfdir=/etc --localstatedir=/var --libdir=lib/x86_64-linux-gnu -Dpython.bytecompile=-1 -Dexamples=false -Dintrospection=true -Dman=true -Dsystemdsystemunitdir=/usr/lib/systemd/system -Dtests=true -Dgtk_doc=true -Dsession_tracking=libsystemd-login
The Meson build system
Version: 1.3.1
Source dir: /builds/bluca/polkit/debian/output/source_dir Build dir: /builds/bluca/polkit/debian/output/source_dir/obj-x86_64-linux-gnu Build type: native build
Project name: polkit
Project version: 124

<...>

Run-time dependency libsystemd found: YES 255
Checking for function "sd_uid_get_display" with dependency libsystemd: YES
Checking for function "sd_pidfd_get_session" with dependency libsystemd: YES
../meson.build:222:37: ERROR: Unknown variable "systemd_dep".
```

Follow-up for 24f1e0af3f7bd17e220cb96201f3c654e737ad34

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -191,16 +191,6 @@ enable_logind = (session_tracking != 'ConsoleKit')
 if enable_logind
   if session_tracking == 'logind'
     logind_dep = dependency('libsystemd', not_found_message: 'logind support requested but libsystemd library not found')
-
-    # systemd unit / service files
-    systemd_systemdsystemunitdir = get_option('systemdsystemunitdir')
-    if systemd_systemdsystemunitdir == ''
-      systemd_dep = dependency('systemd', not_found_message: 'systemd required but not found, please provide a valid systemd user unit dir or disable it')
-      # FIXME: systemd.pc file does not use variables with relative paths, so `define_variable` cannot be used
-      systemd_systemdsystemunitdir = systemd_dep.get_variable(pkgconfig: 'systemdsystemunitdir')
-    endif
-
-    systemd_sysusers_dir = systemd_dep.get_variable(pkgconfig: 'sysusers_dir', default_value: '/usr/lib/sysusers.d')
   else
     logind_dep = dependency('libelogind', not_found_message: 'elogind support requested but libelogind library not found')
   endif
@@ -212,19 +202,25 @@ if enable_logind
 endif
 config_data.set('HAVE_LIBSYSTEMD', enable_logind)
 
-systemd_dep = dependency('systemd').found()
-if systemd_dep
-  tmpfiles_conf = files(
-    'data/polkit-tmpfiles.conf'
-  )
-
-  install_data(
-    tmpfiles_conf,
-    install_dir: pk_prefix / 'lib/tmpfiles.d/'
-  )
-endif
-
 config_data.set('HAVE_PIDFD_OPEN', cc.get_define('SYS_pidfd_open', prefix: '#include <sys/syscall.h>') != '')
+
+# systemd unit / sysuser / tmpfiles.d file installation directories
+systemdsystemunitdir = get_option('systemdsystemunitdir')
+systemd_dep = dependency('systemd', required : false)
+if systemd_dep.found()
+  if systemdsystemunitdir == ''
+    # FIXME: systemd.pc file does not use variables with relative paths, so `define_variable` cannot be used
+    systemdsystemunitdir = systemd_dep.get_variable(pkgconfig : 'systemdsystemunitdir')
+  endif
+  sysusers_dir = systemd_dep.get_variable(pkgconfig : 'sysusers_dir')
+  tmpfiles_dir = systemd_dep.get_variable(pkgconfig : 'tmpfiles_dir')
+else
+  if systemdsystemunitdir == ''
+    systemdsystemunitdir = '/usr/lib/systemd/system'
+  endif
+  sysusers_dir = '/usr/lib/sysusers.d'
+  tmpfiles_dir = '/usr/lib/tmpfiles.d'
+endif
 
 # User for running polkitd
 polkitd_user = get_option('polkitd_user')
@@ -409,12 +405,9 @@ output += '        introspection:            ' + enable_introspection.to_string(
 output += '        Distribution/OS:          ' + os_type + '\n'
 output += '        Authentication framework: ' + auth_fw + '\n'
 output += '        Session tracking:         ' + session_tracking + '\n'
-if session_tracking == 'logind'
-  output += '        systemdsystemunitdir:     ' + systemd_systemdsystemunitdir + '\n'
-endif
-if systemd_dep
-  output += '        systemdtmpfilesdir:       ' + pk_prefix + '/lib/tmpfiles.d' + '\n'
-endif
+output += '        system unit directory:    ' + systemdsystemunitdir + '\n'
+output += '        sysusers.d directory:     ' + sysusers_dir + '\n'
+output += '        tmpfiles.d directory:     ' + tmpfiles_dir + '\n'
 
 output += '        polkitd user:             ' + polkitd_user + ' \n'
 if polkitd_uid != '-'
--- a/data/meson.build
+++ b/data/meson.build
@@ -30,25 +30,28 @@ if enable_pam
   )
 endif
 
-if session_tracking == 'logind'
-  configure_file(
-    input: 'polkit.service.in',
-    output: '@BASENAME@',
-    configuration: service_conf,
-    install: true,
-    install_dir: systemd_systemdsystemunitdir,
-  )
+configure_file(
+  input: 'polkit.service.in',
+  output: '@BASENAME@',
+  configuration: service_conf,
+  install: true,
+  install_dir: systemdsystemunitdir,
+)
 
-  configure_file(
-    input: 'polkit.conf.in',
-    output: '@BASENAME@',
-    configuration: service_conf,
-    install: true,
-    install_dir: systemd_sysusers_dir,
-  )
-endif
+configure_file(
+  input: 'polkit.conf.in',
+  output: '@BASENAME@',
+  configuration: service_conf,
+  install: true,
+  install_dir: sysusers_dir,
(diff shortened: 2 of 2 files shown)


meson.build
```diff
@@ -212,9 +212,9 @@ if enable_logind
   config_h.set10('HAVE_' + func.to_upper(), cc.has_function(func, dependencies: logind_dep))
 
   # systemd unit / service files
+  systemd_dep = dependency('systemd', not_found_message: 'systemd required but not found, please provide a valid systemd user unit dir or disable it')
```
thesamesam, line 215: This change breaks using elogind (https://bugs.gentoo.org/922458).

Need to keep the session_tracking conditional, I think.

thesamesam: I applied this on top which seems to work:
```diff
--- a/meson.build
+++ b/meson.build
@@ -212,11 +212,14 @@ if enable_logind
   config_h.set10('HAVE_' + func.to_upper(), cc.has_function(func, dependencies: logind_dep))
 
   # systemd unit / service files
-  systemd_dep = dependency('systemd', not_found_message: 'systemd required but not found, please provide a valid systemd user unit dir or disable it')
   systemd_systemdsystemunitdir = get_option('systemdsystemunitdir')
-  if systemd_systemdsystemunitdir == '' and session_tracking == 'libsystemd-login'
-    # FIXME: systemd.pc file does not use variables with relative paths, so `define_variable` cannot be used
-    systemd_systemdsystemunitdir = systemd_dep.get_pkgconfig_variable('systemdsystemunitdir')
+  if session_tracking == 'libsystemd-login'
+    systemd_dep = dependency('systemd', not_found_message: 'systemd required but not found, please provide a valid systemd user unit dir or disable it')
+
+    if systemd_systemdsystemunitdir == ''
+      # FIXME: systemd.pc file does not use variables with relative paths, so `define_variable` cannot be used
+      systemd_systemdsystemunitdir = systemd_dep.get_pkgconfig_variable('systemdsystemunitdir')
+    endif
   endif
 
   systemd_sysusers_dir = systemd_dep.get_pkgconfig_variable('sysusers_dir', default: '/usr/lib/sysusers.d')
-- 
2.43.0
```

bluca: No that suggested change is not right, as the sysuser file is needed independently of logind

thesamesam: Sure, but the logic which as-is only installs the sysuser file with systemd anyway.

bluca: Good point, that needs to be fixed too, as it shouldn't and it needs to always be installed

meson.build
```diff
@@ -222,6 +212,16 @@ config_h.set('HAVE_LIBSYSTEMD', enable_logind)
...
 
+# systemd unit / sysuser file installation directories
+systemd_dep = dependency('systemd', not_found_message: 'systemd.pc pkg-config file required but not found')
+systemd_systemdsystemunitdir = get_option('systemdsystemunitdir')
+if systemd_systemdsystemunitdir == ''
+  # FIXME: systemd.pc file does not use variables with relative paths, so `define_variable` cannot be used
+  systemd_systemdsystemunitdir = systemd_dep.get_pkgconfig_variable('systemdsystemunitdir')
+endif
+
+systemd_sysusers_dir = systemd_dep.get_pkgconfig_variable('sysusers_dir', default: '/usr/lib/sysusers.d')
```
alyssais, line 223: This should at least provide an option to manually specify the directory like for system units, in case there is no libsystemd.

bluca: This is unrelated to libsystemd, it's a separate and independent pkg-config file

alyssais: Ah, then in case there's no systemd.pc.

bluca: Then it's a bug in the package/build script/etc and it needs to be provided

alyssais: It seems unnecessarily unkind to packagers to add a new required dependency just to get a single path.

bluca: It's two, and one day it will be three, and then four, and then... - the point of pkg-config is exactly to avoid having to proliferate hundreds of build options in every single build system for every single project, and instead have one interface that can be used to customize things.
And it's for the benefits of the antis too, btw: it's just a text file with a common format, so they can just define it with `systemdsystemunitdir=/dev/null` or something like that et voila, these config files disappear automagically. win/win

alyssais: What about operating systems that systemd doesn't even run on but polkit does, like FreeBSD?  They have to create a systemd.pc file to configure the installation of a bunch of files that they're never going to use, rather than just having a way to not install these files that are irrelevant to them in the first place?

bluca: Yes - again, that's something done _once_ that will work everywhere, for every project being built, instead of having to go around and ask to merge patches to add individual build options to each and every project, and then use them. How is this not better in every way? And if the answer is "because it has 'systemd' in the file name" then I'm going to yawn so hard I'll dislocate my jaw

alyssais: It's nothing to do with it being systemd — I like systemd, I've contributed to systemd, and I'd feel the same way for every other optional dependency.  I think it's uncharitable to assume that disagreement about the best way to handle optional features is some sort of bad faith "I don't like systemd" trolling.

I don't think it's better because it creates work where there could be none (just disabling the feature or using a default path if systemd.pc was not found, for example).  You've even already supplied a default value, but just written the code in a way so that it fails if the .pc file is present before it gets to the point where it could fall back to the default you've already supplied.  How is that better?

bluca: It's not true that it creates more work, it removes it. The alternative is doing more work to add one build option for each path, which is going to be different for every project, find it, and use it. Versus just defining the stub .pc once and using it everywhere.

bluca: Again the important bit is that this is not just about polkit, pretty much everything is starting to use sysusers and tmpfiles.d. Patching N projects to add 2 build options each does not scale well. Defining and installing a stub .pc is done once and works everywhere.

alyssais: But why is it acceptable to fall back to the default value if the .pc file is present but doesn't contain the required option, but it's not acceptable to fall back to the default value if the .pc file is not present?

jrybar-rh: I don't mean to ad fuel to the fire but, TBH, configuring <dependency> if <dependency> is not (and probably never will be) present on the system (be it systemd or whatever) seems like abomination to me (/me just shot himself in foot, probably), Even though we use prepared and good-looking system like pkgconf. @bluca, this is nothing personal, I just believe there are cleaner ways than dropping unused config files in possibly-yet-non-existent directories on systems that have right to go different ways.

classabbyamp: Please don't force a stub .pc file. The existence of `systemd.pc` (even a stub) can imply that systemd exists on the system, and there will be programs that will make that assumption and configure themselves incorrectly.

bluca: > I don't mean to ad fuel to the fire but, TBH, configuring 'dependency' if 'dependency' is not (and probably never will be) present on the system (be it systemd or whatever) seems like abomination to me (/me just shot himself in foot, probably), even though we use prepared and good-looking system like pkgconf. @bluca, this is nothing personal, I just believe there are cleaner ways than dropping unused config files in possibly-yet-non-existent directories on systems that have right to go different ways. But maybe I'm still a purist in this topic :D

I disagree, but it's your project, so as you wish - the pkgconf is now optional, and the well-known default installation directories are used as a fallback if missing.

meson.build
```diff
@@ -212,20 +202,26 @@ if enable_logind
...
-  )
+config_data.set('HAVE_PIDFD_OPEN', cc.get_define('SYS_pidfd_open', prefix: '#include <sys/syscall.h>') != '')
 
-  install_data(
-    tmpfiles_conf,
-    install_dir: pk_prefix / 'lib/tmpfiles.d/'
-  )
+# systemd unit / sysuser / tmpfiles.d file installation directories
+systemdsystemunitdir = get_option('systemdsystemunitdir')
+systemd_dep = dependency('systemd')
```
classabbyamp, line 209:
```suggestion
systemd_dep = dependency('systemd', required : false)
```
this is necessary to build without systemd

otherwise, this works well and successfully builds and installs in void's package build system with `-Dintrospection=true  -Dman=true  -Dsession_tracking=elogind -Dpam_prefix=/usr/lib/pam.d`

jrybar-rh:
I got a message that tools like tmpfiles.d and systemd-sysusers can also exist separately from systemd, hence it make sense (since we use those tools) to provide configuration files for them. **IF** they are not desired on a specific distro, a package maintainer for that distro can always exclude those files. Agreed?  
Anyway, thanks, @bluca, for your time and your work!

bluca:
> I got a message that tools like tmpfiles.d and systemd-sysusers can also exist separately from systemd, hence it make sense (since we use those tools) to provide configuration files for them

Yes there are separate implementations in bash for both. Thanks.

PR #425, opened 2024-02-16
doc: document --no-debug, which is set by default in our own systemd …

### LINKED ISSUE #154: Documentation about --no-debug mode unclear
In gitlab.freedesktop.org by jrybar on Aug 31, 2021, 17:07

Link to the original issue: https://gitlab.freedesktop.org/polkit/polkit/-/issues/153
AFAIK there is no mention of --no-debug mode in polkitd's manpage (actually no option is described there).  
This may cause confusion when --no-debug mode is activated and all output goes to /dev/null without any mention in journal.

lf-:
…service

This can be immensely frustrating to users since there is no mention whatsoever in the documentation of why their logs are getting eaten.

Fixes #154.

### DIFF
--- a/docs/man/polkit.xml
+++ b/docs/man/polkit.xml
@@ -666,8 +666,18 @@ polkit.Result = {
         with the JavaScript filename and line number. Log entries are
         emitted using the <constant>LOG_AUTHPRIV</constant> flag meaning
         that the log entries usually ends up in the file
-        <filename>/var/log/secure</filename>. The
-        <function>log()</function> method is usually only used when
+        <filename>/var/log/secure</filename> or in
+        <command>journalctl -b SYSLOG_FACILITY=10</command> on systemd.
+      </para>
+      <para>
+        On some distributions, the flag <literal>--no-debug</literal> is passed
+        to <command>polkitd</command> by default, discarding log output. This
+        means that to debug a rule, it must removed first, say with
+        <command>systemctl edit --runtime polkit.service</command>, for the
+        current boot.
+      </para>
+      <para>
+        The <function>log()</function> method is usually only used when
         debugging rules. The <type>Action</type> and
         <type>Subject</type> types has suitable
         <function>toString()</function> methods defined for easy
--- a/docs/man/polkitd.xml
+++ b/docs/man/polkitd.xml
@@ -24,6 +24,9 @@
   <refsynopsisdiv>
     <cmdsynopsis>
       <command>polkitd</command>
+      <arg><option>--replace|-r</option></arg>
+      <arg><option>--no-debug|-n</option></arg>
+      <arg><option>--help|-h</option></arg>
     </cmdsynopsis>
   </refsynopsisdiv>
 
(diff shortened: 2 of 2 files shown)


docs/man/polkit.xml
```diff
@@ -666,8 +666,18 @@ polkit.Result = {
         with the JavaScript filename and line number. Log entries are
         emitted using the <constant>LOG_AUTHPRIV</constant> flag meaning
         that the log entries usually ends up in the file
-        <filename>/var/log/secure</filename>. The
-        <function>log()</function> method is usually only used when
+        <filename>/var/log/secure</filename> or in
+        <command>journalctl -b SYSLOG_FACILITY=10</command> on systemd.
+      </para>
+      <para>
+        On some distributions, the flag <literal>--no-debug</literal> is passed
```
mbiebl, line 673: Hm, this is what data/polkit.service.in provides by default
https://github.com/polkit-org/polkit/blob/main/data/polkit.service.in#L11

So I find it a bit misleading to call this something distro specific.

Fwiw, I would like to see the `--no-debug` flag dropped and replaced with `--debug`.
The default behavious would be a reasonable non-verbose/non-debug output.
And if you want to have explicit debug output (say for a bug report), one should use `--debug`.

bluca: When everything works, system services should be silent. We should however add support for LogControl1 to dynamically change the level, I started a branch somewhere but ENOTIME

mbiebl: Well, polkitd does produce log messages, but no debug log messages, e.g. I get
```
# journalctl -u polkitd.service
Feb 25 09:16:59 mars polkitd[993]: Started polkitd version 124
Feb 25 09:16:59 mars polkitd[993]: Loading rules from directory /etc/polkit-1/rules.d
Feb 25 09:16:59 mars polkitd[993]: Loading rules from directory /usr/share/polkit-1/rules.d
Feb 25 09:16:59 mars polkitd[993]: Finished loading, compiling and executing 11 rules
Feb 25 09:16:59 mars polkitd[993]: Acquired the name org.freedesktop.PolicyKit1 on the system bus
Feb 25 09:16:59 mars systemd[1]: Started polkit.service - Authorization Manager.
```

I'm not convinced we should dump debug log messages into the journal unless explicitly requested.

lf-: > When everything works, system services should be silent. We should however add support for LogControl1 to dynamically change the level, I started a branch somewhere but ENOTIME

This isn't my view but I understand this view. :)

jrybar-rh: General information about polkit doesn't seem the right place for information about some binary's options, usage of which is distro-specific, no matter how many distros use it. I like switching polkitd to silent by default though and turning verbose on explicitly. I'll incorporate this idea into my plans for implementing log levels.

docs/man/polkitd.xml
```diff
@@ -40,6 +43,12 @@
       whenever an application calls into the service.
     </para>
 
+    <para>
+      Passing the <option>--no-debug</option> to <command>polkitd</command>
+      will discard the messages from usages of <literal>polkit.log()</literal>
```
jrybar-rh, line 48: "from usages of <literal>polkit.log()</literal>" is an unnecessary detail. Users do no need to know the code details.
Just saying "Do not pass debug information to logs." is sufficient and targeted to wider audience.

jrybar-rh:
@lf-  Friendly ping. Is there still a desire to have this merged after post-review fixes? I like the extension of the polkitd manpage.

jrybar-rh:
Closing this PR for no activity and outdated changes. '--no-debug' will be properly documented in 'polkitd' manpage if actually kept (WIP).

PR #427, opened 2024-02-27
pkexec: fixed different quotes in messages

TheUtopian:


jrybar-rh:
Hello and thank you for your time and enthusiasm.  
I see two main issues in this PR, as  
1. it breaks the alignment with the rest of the programs distributed in ./src/programs/ subdirectory, where all the msg strings adhere to the standard (one of many) of ``'` quotes, as long as with the rest of the code in polkit, and  
2. no justification for the change was provided in the commit nor PR message. Why has this change been made and why just this one source file?  

For these reasons I'll close this PR. I believe that if there is a strong reason to make this change, I suggest it is done consistently throughout the whole project.  

Many thanks again, stay tuned.

TheUtopian:
Hello @jrybar-rh, sorry that I didn't provide any justification, I thought it's just a typos in the file. I made this PR because it looks really bad when we have different quotes in the message.
Screenshot_2024-03-07-19-43-52_1920x1080
I chose this ( **'** ) quote instead of this ( **`** ) because it's more clear/readable in small displays.

I didn't know this quotes were used in other files. If you will agree, I could replace them in other files as well?

PR #430, opened 2024-03-10
meson.build: allow overriding sysusers.d dir

tmuehlbacher:
fixes builds for non-FHS systems like Nix.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -219,7 +219,10 @@ if enable_logind
     systemd_systemdsystemunitdir = systemd_dep.get_pkgconfig_variable('systemdsystemunitdir')
   endif
 
-  systemd_sysusers_dir = systemd_dep.get_pkgconfig_variable('sysusers_dir', default: '/usr/lib/sysusers.d')
+  systemd_sysusers_dir = get_option('systemdsysusersdir')
+  if systemd_sysusers_dir == '' and session_tracking == 'libsystemd-login'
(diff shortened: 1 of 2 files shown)


tmuehlbacher:
> This is read from pkgconf exactly for this use case, so that distribution can ship a custom pc file with their preferred locations, what's the reason why nix is not using that?

In Nix, the resulting files from a build are installed into one (or more) unique store paths that includes the hash of the package and all of its build inputs/parameters. So systemd cannot know ahead of time where polkit will have to install its outputs to.

Some examples paths:
```
/nix/store/jbc0ga700chv62qr7q5x28bxrpmxk04g-polkit-124/lib/sysusers.d/polkit.conf
/nix/store/n85xhkvxv8vhbshyqwwk5l5aiq4xldcc-systemd-minimal-255.2/share/polkit-1/actions/org.freedesktop.systemd1.policy
```

I've specifically implemented this change to be similar to the already existing `systemdsystemunitdir` meson option. There technically is still a programming error in the meson file because `systemd_dep` is defined inside of an `if` condition but then used unconditionally in this code path. With this fix you can at least work around this error by specifying both the system unit dir and the sysusers dir as meson options.

bluca:
This is how cross compilation already does it btw, and it works just fine for that case, so I am pretty sure you can do the same in nix too.

tmuehlbacher:
Ok, using a different approach now, see https://github.com/NixOS/nixpkgs/pull/295087
I am closing this PR as it seems that you already have a separate one to fix the build failure with https://github.com/polkit-org/polkit/pull/417

PR #442, opened 2024-04-04
tmpfiles.d: install and use systemd-tmpfiles.d mechanism to .packit conf

vmihalko:
## Summary
tmpfiles.d will automatically create the /etc/polkit-1/rules.d directory (and any non-existing directories on the path). After reboot, the directory contents will remain unchanged, and after removing the directory & rebooting, an empty directory will be created.

tmpfiles.d in fedora: https://docs.fedoraproject.org/en-US/packaging-guidelines/Tmpfiles.d/
tmpfiles.d manpage: https://www.freedesktop.org/software/systemd/man/latest/tmpfiles.d.html

### DIFF
--- a/.packit/polkit.spec
+++ b/.packit/polkit.spec
@@ -137,6 +137,7 @@ rm -f $RPM_BUILD_ROOT%{_libdir}/*.la
 %{_bindir}/pkttyagent
 %dir %{_prefix}/lib/polkit-1
 %{_prefix}/lib/polkit-1/polkitd
+%{_tmpfilesdir}/polkit-tmpfiles.conf
 
 # see upstream docs for why these permissions are necessary
 %attr(4755,root,root) %{_bindir}/pkexec
--- a/data/polkit-tmpfiles.conf
+++ b/data/polkit-tmpfiles.conf
@@ -1,0 +1,1 @@
+d /etc/polkit-1/rules.d 0750 root polkitd - -
--- a/meson.build
+++ b/meson.build
@@ -220,6 +220,18 @@ if enable_logind
 endif
 config_h.set('HAVE_LIBSYSTEMD', enable_logind)
 
+systemd_dep = dependency('systemd').found()
+if systemd_dep
+  tmpfiles_conf = files(
+    'data/polkit-tmpfiles.conf'
+  )
+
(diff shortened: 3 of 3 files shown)


jtojnar:
What is the purpose of automatically creating `/etc` subdirectory? Should not it be unnecessary unless the administrator wants to configure polkit, in which case they can just create it themselves?

jtojnar:
Also the `sysusers.d` will create a group matching the `-Dpolkitd_user` option so hardcoding it to `polkitd` will break when the packager uses a different one. For example, on NixOS, we use `polkituser`.

jrybar-rh:
It all is on purpose.
The tmpfiles.d is used to create polkit subdirectory in `/etc`, because the directory needs specific ownership mode, it needs to be owned by root and only readable by root and `polkitd` group. This is for security reasons, so that polkit daemon cannot change .rules files in case of getting hijacked. It should have as minimal privileges as possible. The subdirectory is pre-created so no administrator omits to set the appropriate mode and ownership.  
The group is created for the reasons above. `@polkitd_uid` can be set during meson configuration. There is a habit (or standard, I'll verify later) that the user name is the same as for the daemon.

jtojnar:
> The tmpfiles.d is used to create polkit subdirectory in `/etc`, because the directory needs specific ownership mode, it needs to be owned by root and only readable by root and `polkitd` group. This is for security reasons, so that polkit daemon cannot change .rules files in case of getting hijacked. It should have as minimal privileges as possible. The subdirectory is pre-created so no administrator omits to set the appropriate mode and ownership.

Thanks for the explanation. That makes sense, though it should probably be documented somewhere. Maybe a comment in the `tmpfiles.d` file?

> The group is created for the reasons above. `@polkitd_uid` can be set during meson configuration. There is a habit (or standard, I'll verify later) that the user name is the same as for the daemon.

Yes, `sysusers.d` [enforces this convention](https://www.freedesktop.org/software/systemd/man/latest/sysusers.d.html#u). However, the issue is that this file hardcodes the group name so it will not be correct if packager changes it with `-Dpolkitd_user`.

I have opened https://github.com/polkit-org/polkit/pull/555 fixing both of those issues.

PR #443, opened 2024-04-15
Improve libs-only option

michaelolbrich:
## Summary
This enables the polkit-agent-1 library for libs-only builds and skips the expat dependency when it is not needed.


michaelolbrich:
The test failure seems unrelated:
```
  ../meson.build:1:0: ERROR: Meson version is 0.61.2 but project requires >= 0.63.0
```
This pull request does not change the meson version.

PR #448, opened 2024-05-07
Macro housekeeping

notpeelz:
1. I removed feature test macro definitions (i.e `_GNU_SOURCE`/`_XOPEN_SOURCE`) from polkitbackend because `_GNU_SOURCE` is already defined in the root meson.build.
Explicitly defining `_XOPEN_SOURCE` is redundant because `_GNU_SOURCE` implies `_XOPEN_SOURCE 700` (according to the comments in glibc's `features.h`)

2. I simplified the handling of `config.h`; instead of having `#include "config.h"` in every file, we can use the compiler's `-include` option to do that for us
I had to bump the meson version requirement to >=1.4.0 to use the `file.full_path()` API.

3. I moved the `-std=c99` flag to the project's `default_options`, as it appears to have been fixed (https://github.com/mesonbuild/meson/pull/10170)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -5,9 +5,10 @@ project(
   default_options: [
     'buildtype=debugoptimized',
     'prefix=/usr',
+    'c_std=c99',
     'cpp_std=c++17',
   ],
-  meson_version: '>= 0.63.0',
+  meson_version: '>= 1.4.0',
 )
 
 pk_version = meson.project_version()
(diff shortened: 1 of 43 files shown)


vmihalko:
Hello, thank you for your contributions! Could you please create a separate PR for each change (item in your description)?

notpeelz:
I initially had them as separate commits, but decided to squash them because the changes are tightly coupled.
I'll try to split them into stacked PRs.

notpeelz:
Change `#2` split into #454
Change `#1` split into #455 (depends on #454)

I'll make a new PR for change `#3` once #454 is merged, so to avoid unnecessary merge conflicts.

PR #454, opened 2024-05-17
Refactor config.h

notpeelz:
I simplified the handling of `config.h`.
Instead of having `#include "config.h"` in every file, we can use the compiler's `-include` option to do that for us.

I had to bump the meson version requirement to >=1.4.0 to use the `file.full_path()` API.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -64,7 +64,7 @@ top_inc = include_directories('.')
 
 cc = meson.get_compiler('c')
 
-config_h = configuration_data()
+config_data = configuration_data()
 
 # defines
 set_defines = [
@@ -79,11 +79,11 @@ set_defines = [
 ]
 
 foreach define: set_defines
-  config_h.set_quoted(define[0], define[1])
+  config_data.set_quoted(define[0], define[1])
 endforeach
 
 # Globally define_GNU_SOURCE and therefore enable the GNU extensions
-config_h.set('_GNU_SOURCE', true)
+config_data.set('_GNU_SOURCE', true)
 
 # functions
 check_functions = [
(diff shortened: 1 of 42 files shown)


meson.build
```diff
@@ -355,6 +351,16 @@ if enable_introspection
...
 
+config_h = configure_file(
+  output: 'config.h',
+  configuration: config_data,
+)
+
+compiler_common_flags += ['-include', config_h.full_path()]
+
+add_project_arguments(compiler_common_flags + compiler_c_flags, language: 'c')
+add_project_arguments(compiler_common_flags + compiler_cpp_flags, language: 'cpp')
```
mrc0mmand, line 362: I think this can be simplified a bit:
 - there's no C++ code, so the second `add_project_arguments()` invocation can be dropped
 - no need to keep the CFLAGS in a separate array, just use `add_project_arguments()` directly, since it appends the argument(s)
 - you don't need to use the full path to the `config.h` file, so the related meson version bump can be dropped

```suggestion
add_project_arguments(['-include', 'config.h'], language: 'c')
```

src/programs/pkexec.c
```diff
@@ -201,7 +197,7 @@ open_session (const gchar *user_to_auth,
       for (n = 0; envlist[n]; n++)
 	{
 	  const char *envitem = envlist[n];
-
```
mrc0mmand, line 204: Please split the whitespace changes into a separate commit.

mrc0mmand (approved):

meson.build
mrc0mmand: Apologies, I completely missed the fact that the `compiler_c_flags` stuff is preexisting (and that there's at least one cpp file). In that case ignore the previous two points.

As for the last one, I think the meson bump is too big for a very little gain, but I'll leave the decision to @jrybar-rh.

jrybar-rh: Hello, IMO there is a possibility of regrets when we sacrifice CI for F39. I know, just 5 months, but if sith happens, testing fixes for F39 would be unnecessarily difficult. Can we make the 'bump-needing' feature postponed (and reminded by a new issue)?

notpeelz: Okay, I'll make a new PR for it once this is merged.

notpeelz: See #465

PR #455, opened 2024-05-17
Fix duplicate feature test macros

notpeelz:
I removed feature test macro definitions (i.e `_GNU_SOURCE`/`_XOPEN_SOURCE`) from polkitbackend because `_GNU_SOURCE` is already defined in the root meson.build.
Explicitly defining `_XOPEN_SOURCE` is redundant because `_GNU_SOURCE` implies `_XOPEN_SOURCE 700` (according to the comments in glibc's `features.h`)

Depends on #454

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -83,9 +83,6 @@ foreach define: set_defines
   config_data.set_quoted(define[0], define[1])
 endforeach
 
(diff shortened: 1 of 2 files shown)


notpeelz:
@jrybar-rh Now that all the other PRs are merged, this should be ready for review.

PR #461, opened 2024-06-10
test: drop mocklibc

mrc0mmand:
Let's get rid of mocklibc and replace it with a simple combination of mount & user namespaces + bind mount to replace the host's /etc with our own version.

This means we don't $LD_PRELOAD the mocklibc DSO, but instead run each unit test through a very simple python wrapper that sets up a temporary user & mount namespace through the unshare() syscall,  gains "fake" root using uid_map and gid_map, overmounts /etc in this new namespace (with our own custom test files), and then executes the test binary itself. Check user_namespaces(7) for more information about the namespace shenanigans.


mrc0mmand:
@jrybar-rh here's the promised mocklibc replacement. It replaces mocklibc with a simple python script that sets up a user and mount namespace, so we can overmount `/etc` for the unit tests without disturbing the underlying system (and so we can do that at all as an unprivileged user). This also allows us to run all unit tests under sanitizers, which uncovered one memory leak, see the second commit. The last commit also shows that the tests can be now run as part of an RPM build if desired.

While testing this I noticed the CIs (and my local runs) skip the netgroup tests, because the test sources don't include `config.h` which necessary `#define`s, so preferably https://github.com/polkit-org/polkit/pull/454 should land first which mitigates this issue without having to scatter more `#include "config.h"` across the codebase.

test/polkit/meson.build
```diff
@@ -20,8 +20,8 @@ foreach test_unit: test_units
 
   test(
     test_unit,
-    exe,
-    env: test_env,
-    is_parallel: false,
+    test_wrapper,
+    args: ['--data-dir', test_data_dir, exe.full_path()],
```
jrybar-rh, line 24: Is this applicable to still-supported F39, which we use for CI?

mrc0mmand: Yes, thanks to the last patch the tests run in all Packit CIs, including the F39 one, and everything seems to pass there.

PR #465, opened 2024-06-26
Dont hardcode path to config.h

notpeelz:
Split from #454 as requested.

This is a minor change that requires meson >= 1.4.0, which isn't available in Fedora 39.
F39 becomes EOL on 2024-11-12.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -375,12 +375,12 @@ if gettext
   config_data.set('ENABLE_GETTEXT', 1)
 endif
 
-configure_file(
+config_h = configure_file(
   output: 'config.h',
   configuration: config_data,
 )
 
-compiler_common_flags += ['-include', 'config.h']
+compiler_common_flags += ['-include', config_h.full_path()]
 
 add_project_arguments(compiler_common_flags + compiler_c_flags, language: 'c')
 add_project_arguments(compiler_common_flags + compiler_cpp_flags, language: 'cpp')
--- a/meson.build
+++ b/meson.build
@@ -8,7 +8,7 @@ project(
     'c_std=c99',
     'cpp_std=c++17',
   ],
-  meson_version: '>= 0.63.0',
+  meson_version: '>= 1.4.0',
 )
 
 pk_version = meson.project_version()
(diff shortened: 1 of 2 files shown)


vmihalko:
Could you please resolve the conflicts? Then, we can proceed with the merge.

notpeelz:
Done

notpeelz:
Ah... gotta love eternally-outdated distros.

Feel free to close this PR if you don't think it's worth it.

notpeelz:
Oh that's great news. I've rebased and applied your patch.

jrybar-rh:
Ah, here seems to be the hiccup :smile: 
https://github.com/polkit-org/polkit/actions/runs/13275796335/job/37065055620

mrc0mmand:
> Ah, here seems to be the hiccup 😄 https://github.com/polkit-org/polkit/actions/runs/13275796335/job/37065055620

Ah, whoops, forgot about that one. I'll submit a follow-up.

mrc0mmand:
> > Ah, here seems to be the hiccup 😄 https://github.com/polkit-org/polkit/actions/runs/13275796335/job/37065055620
> 
> Ah, whoops, forgot about that one. I'll submit a follow-up.

-> https://github.com/polkit-org/polkit/pull/551

meson.build
```diff
@@ -375,12 +375,12 @@ if gettext
...
 endif
 
-configure_file(
+config_h = configure_file(
   output: 'config.h',
   configuration: config_data,
 )
 
-compiler_common_flags += ['-include', 'config.h']
+compiler_common_flags += ['-include', config_h.full_path()]
```
eli-schwartz, line 383: You don't need to bump the minimum required version. You do "need" the full path if polkit can ever be built as a subproject -- the compiler is always run from the global_build_root(), that is, the same directory as the `build.ninja` file, and config.h is generated where current_build_dir() == project_build_root(), which is the same directory as `build.ninja` except in a subproject.

But regardless, the string `config_h.full_path()` is provided for API equivalence with other objects. You could have always reliably guaranteed the location quite simply:

```meson
config_h_full_path = meson.current_build_dir() / 'config.h'
```

No version bump needed.

mrc0mmand: Ah, that's actually pretty neat, thanks a lot for the explanation!

PR #471, opened 2024-07-02
meson: do not depend on systemd

vmihalko:


bluca:
@jrybar-rh this is superseded and can be closed

vmihalko:
Agree.

PR #482, opened 2024-08-06
Add --log-level and LogControl1 support

bluca:
Alternative to https://github.com/polkit-org/polkit/pull/477

### DIFF
--- a/src/polkitbackend/polkitbackendauthority.c
+++ b/src/polkitbackend/polkitbackendauthority.c
@@ -52,6 +52,7 @@ enum
 };
 
 static guint signals[LAST_SIGNAL] = {0};
+static guint polkit_authority_log_level = LOG_LEVEL_ERROR;
 
 G_DEFINE_ABSTRACT_TYPE (PolkitBackendAuthority, polkit_backend_authority, G_TYPE_OBJECT);
 
@@ -1561,6 +1562,7 @@ _color_get (_Color color)
 
 void
 polkit_backend_authority_log (PolkitBackendAuthority *authority,
+                              const guint message_log_level,
                               const gchar *format,
                               ...)
 {
@@ -1571,13 +1573,18 @@ polkit_backend_authority_log (PolkitBackendAuthority *authority,
   gchar *message;
   va_list var_args;
 
+  if (message_log_level > polkit_authority_log_level)
+  {
+	  return;
+  }
+
   g_return_if_fail (POLKIT_BACKEND_IS_AUTHORITY (authority));
 
   va_start (var_args, format);
   message = g_strdup_vprintf (format, var_args);
   va_end (var_args);
 
-  syslog (LOG_NOTICE, "%s", message);
+  syslog (message_log_level, "%s", message);
 
   g_get_current_time (&now);
   now_time = (time_t) now.tv_sec;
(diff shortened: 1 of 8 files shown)


src/polkitbackend/polkitbackendauthority.c
```diff
@@ -1591,3 +1598,24 @@ polkit_backend_authority_log (PolkitBackendAuthority *authority,
 
   g_free (message);
 }
+
+void
+polkit_backend_authority_set_log_level (const gchar *level)
+  {
+    if (g_strcmp0 (level, "debug") == 0)
```
jrybar-rh, line 1712: This doesn't allow using numerical values for shortopt. Using sth like "polkitd -l crit" doesn't look like a standard solution.

src/polkitbackend/polkitbackendauthority.c
```diff
@@ -1591,3 +1598,24 @@ polkit_backend_authority_log (PolkitBackendAuthority *authority,
...
+      polkit_authority_log_level = (guint) LOG_LEVEL_DEBUG;
+    else if (g_strcmp0 (level, "info") == 0)
+      polkit_authority_log_level = (guint) LOG_LEVEL_INFO;
+    else if (g_strcmp0 (level, "notice") == 0)
+      polkit_authority_log_level = (guint) LOG_LEVEL_NOTICE;
+    else if (g_strcmp0 (level, "warning") == 0)
+      polkit_authority_log_level = (guint) LOG_LEVEL_WARNING;
+    else if (g_strcmp0 (level, "err") == 0)
```
jrybar-rh, line 1720: Some options are abbreviated, some are not. Inconsistency brings confusion.

src/polkitbackend/polkitbackendauthority.c
bluca: I have never seen any project expose this with numeric values? In systemd we do use the string names though, see SYSTEMD_LOG_LEVEL and LogControl1

https://www.freedesktop.org/software/systemd/man/latest/systemd.html#%24SYSTEMD_LOG_LEVEL
https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.LogControl1.html

I don't think it's necessary to expose this option as a numeric value for users, it's something that programmers might understand, but it would be very opaque for a normal user

src/polkitbackend/polkitbackendauthority.c
bluca: Yeah these mixed abbreviations are very strange and I don't understand why they are named like that, however, that's just how syslog levels are called and they are pretty much the standard, and how they are called in the LogControl1 protocol, in all log parsers and viewers like journald and journalctl (ie: you do journalctl -p err rather than -p error) - I think consistency between ways to configure and use this and across the ecosystem is important here, more than aesthetic on an individual project level, although if I could go back in time I'd fix them to be consistently named:

https://man7.org/linux/man-pages/man3/syslog.3.html
https://www.freedesktop.org/software/systemd/man/latest/systemd.html#%24SYSTEMD_LOG_LEVEL
https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.LogControl1.html

src/polkitbackend/polkitbackendauthority.c
jrybar-rh: Sorry, but using numerical values would be confusing (for administrators, mind you), but using sometimes shortened words for shortopts wouldn't?
This is not an intention to massage my ego, but it simply doesn't make sense or consistency.
I've seen many standard programs that use shortopt with a number and longopt with a string.
If polkitd wasn't such a small program expected to be used frequently from shell directly, I'd be for creating one GOption "-l" with nums and another one for "--log-levels" for textual parameters, which would be consistently uncut.

bluca: I meant for log levels. I don't remember anything common enough that takes parameters for the log level as an opaque number. The enum is an implementation detail of syslog after all, the way it's exposed to users is via these strings

src/polkitbackend/polkitbackendauthority.c
jrybar-rh: But the syslog manpage, from which LogControl got apparently inspired, communicates on the programmer level, not on the level of UI. Deriving textual parameters used directly by user from program's enum seems like a UI mistake made long ago. Do we want to continue in this evil? Who do we make this for? For admins who would be confused by numbers but not by half-words derived from enums?
I say we should use whole words for UI, program internals regardless, while we're at it.

bluca: It needs to be the same yes, because again that's what user interfaces like systemctl and journalctl expose, so it's worse for usability to diverge: https://www.freedesktop.org/software/systemd/man/latest/systemctl.html#service-log-level%20SERVICE%20%5BLEVEL%5D

What we can do, is accept both versions as input, as aliases. That's easy to do, I'll push shortly.

PR #485, opened 2024-08-08
test: skip if the namespace cannot be set up due to permissions or if python3 is too old

bluca:
On some build systems, like Debian's, the tests do not have permissions to create new namespaces, so skip gracefully in that case

Follow-up for 8e17f09c770bc2efd5deb40ba2b6032d40603578

### DIFF
--- a/test/wrapper.py
+++ b/test/wrapper.py
@@ -32,6 +32,9 @@ def setup_test_namespace(data_dir):
     except PermissionError:
         print("Lacking permissions to set up test harness, skipping")
         sys.exit(77)
+    except AttributeError:
+        print("Python 3.12 is required for os.unshare(), skipping")
+        sys.exit(77)
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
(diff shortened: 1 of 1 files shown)


jrybar-rh:
@mrc0mmand Skip or fail the test if the user doesn't have permission to run it?

bluca:
Let's not fail please, as it breaks package builds. Hit this issue after uploading the new version to Debian. We skip tests that require privileges in systemd too.

test/wrapper.py
```diff
@@ -17,17 +17,24 @@ def setup_test_namespace(data_dir):
...
+        with open("/proc/self/gid_map", "w") as f:
+            f.write(f"0 {egid} 1")
+
+        # Overmount /etc with our own version
+        subprocess.check_call(["mount", "--bind", os.path.join(data_dir, "etc"), "/etc"])
+    except PermissionError:
+        print("Lacking permissions to set up test harness, skipping")
+        sys.exit(77)
+    except AttributeError:
+        print("Python3 3.12 is required for os.unshare(), skipping")
```
mrc0mmand, line 36:
```suggestion
        print("Python 3.12 is required for os.unshare(), skipping")
```

PR #486, opened 2024-08-08
Add meson option to use gettext for translations in .policy files

bluca:
Disabled by default. Will be used in Debian/Ubuntu and derivatives,
where this patch has been carried out-of-tree for ~15 years.

Fixes https://github.com/polkit-org/polkit/issues/35

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -342,6 +342,11 @@ if enable_introspection
   dependency('gobject-introspection-1.0', version: '>= 0.6.2')
 endif
 
+gettext = get_option('gettext')
+if gettext
+  config_data.set('ENABLE_GETTEXT', 1)
+endif
+
 configure_file(
   output: 'config.h',
   configuration: config_data,
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -19,3 +19,4 @@ option('introspection', type: 'boolean', value: true, description: 'Enable intro
 
 option('gtk_doc', type: 'boolean', value: false, description: 'use gtk-doc to build documentation')
 option('man', type: 'boolean', value: false, description: 'build manual pages')
+option('gettext', type: 'boolean', value: false, description: 'use gettext for translations')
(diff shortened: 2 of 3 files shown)


zerocool438:
Operating system fedora 40. 
After this patch is installed, the policy prompt is correct for the first time every time the polkit service is restarted. After that, only English is displayed.

zerocool438:
@robert-ancell @bluca

zerocool438:
systemd : 255.10-3.fc40
polkit: 124-2.fc40

zerocool438:
1、sudo systemctl  restart polkit.service
2、timedatectl  set-timezone Asia/Singapore
微信图片_20240823111301

After that,  Execute "timedatectl set-timezone Asia/Singapore" again.
微信图片_20240823111310

zerocool438:
> @zerocool438 thanks for testing - could you please try again with this new version just pushed?
I'll give it a try.

zerocool438:
> @zerocool438 thanks for testing - could you please try again with this new version just pushed?
I noticed that you added judgment to the result value. I can just tell you the result value in these two executions.
source code:
截图 2024-08-26 15-28-27
journal:
截图 2024-08-26 15-25-21

@bluca

bluca:
Is the problem with the GUI pop-up that you showed previously solved with the latest code changes in this PR?

zerocool438:
> Is the problem with the GUI pop-up that you showed previously solved with the latest code changes in this PR?

No, this PR didn't fix it.

zerocool438:
> Is the problem with the GUI pop-up that you showed previously solved with the latest code changes in this PR?

The phenomenon remains the same.

bluca:
Ok thanks, I've switched this to a disabled-by-default meson option. We can enable this in Debian and derivatives, and it can be kept disabled in Fedora and derivatives.

bluca:
@jrybar-rh gentle ping - this is disabled by default so shouldn't affect Fedora, but it would allow us in Debian/Ubuntu to drop a downstream patch that we have been carrying for more than 10 years. Thanks!

v1993:
Should there be some kind of documentation for this feature aside from this very PR existing? It's also not really clear if it should be utilized in practice - apparently only if you're targeting Debian and derivatives? But it's also used by systemd?..

PR #489, opened 2024-08-09
jsauthority: Bump mozjs to 128

xry111:
## Summary

Firefox ESR 128 has been released and GNOME 47 has been updated to use it.  Do the same for Polkit.

`JSErrorReport::filename` is now a `ConstUTF8CharsZ`.  Invoke `c_str()` to get the C string.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
 
       - name: Install build & test dependencies
         run: |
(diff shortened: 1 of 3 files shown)


bluca:
Maybe it's time to drop mozjs? duktape's backend has been available and in use in major distros for a few years now

jrybar-rh:
@bluca Yes, this has been discussed months ago in the Github Discussions and it was mentioned in the message following the last release that 125 is the last release with mozjs.

xry111:
So see #494 to drop it.

PR #496, opened 2024-09-03
build: Don't install daemon files when building libs only

hadess:
Those files are only required when the daemon is in use, and because of their hard-coded paths when libsystemd isn't available, it fails to build in Flatpak.

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -30,28 +30,30 @@ if enable_pam
   )
 endif
 
-configure_file(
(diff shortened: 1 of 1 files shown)


maxrdz:
I know this was already resolved, but just wanted to leave a thank you. Was resolving flatpak issues, and was pulling my hair out as to why passing `-Dlibs-only` was leading to polkit *still* trying to install systemd files.

PR #497, opened 2024-09-05
meson: Support "openembedded" OS for PAM configuration

ycongal-smile:
## Summary
The Yocto/Openembedded project uses PAM integration with `common-*` files instead of the default `system-*`.
This PR adds  a `openembedded` `os_type` in meson (using the `suse` code path) to match that.

In Openembedded, same as Suse/Solaris: PAM files are common-*:
* PAM_FILE_INCLUDE_AUTH: common-auth
* PAM_FILE_INCLUDE_ACCOUNT: common-account
* PAM_FILE_INCLUDE_PASSWORD: common-password
* PAM_FILE_INCLUDE_SESSION: common-session

Since the automated OS detection will not work in a cross-compiler case. Polkit is built using the default `system-*` files and OE need to maintain a patch to change to `common-*`.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -298,7 +298,7 @@ endif
 
 pam_include = get_option('pam_include')
 if pam_include == ''
-  if ['suse', 'solaris'].contains(os_type)
+  if ['suse', 'solaris', 'openembedded'].contains(os_type)
     pam_conf = {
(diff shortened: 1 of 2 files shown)


ycongal-smile:
Thanks!

PR #501, opened 2024-09-16
agent helper: support separate socket-activated service to run without SETUID

bluca:
SETUID binaries are considered harmful, as te execution context is under the control of unprivileged attackers.

Enhance the polkit pam agent helper with a new mode: when running under systemd, add a socket-activated service that the helper will run under, as root. The agent talks to this service via AF_UNIX instead of spawning it, and STDIN/STDOUT are connected as before. The helper can make use of PID FDs and SO_PEERCRED to reliably identify the caller. In order to do this, a third version of the auth D-Bus method is added, that also takes a subject, built using the PID FD.
If the AF_UNIX socket is not present, the agent will fork the helper as before, with no changes.

Fixes https://github.com/polkit-org/polkit/issues/169


bluca:
@jrybar-rh gentle ping

jrybar-rh:
> @jrybar-rh gentle ping

Gentle ack.

Though, I probably won't have time to review this till mid-Jan. But I will try to get to this asap.

src/polkitagent/polkitagenthelper-pam.c
```diff
@@ -113,7 +127,47 @@ main (int argc, char *argv[])
       goto error;
     }
```
jrybar-rh, line 129: Hi Luca, formally it all looks good, thanks for the hard work! Just to clear out the purpose, we still have a SUID binary (because of pam stack), but at least it runs in systemd-sandbox, right?

jrybar-rh: Socket activation is manual on purpose, or `WantedBy` is missing?

bluca: It's manual on purpose on the service, as that is only triggered by the socket unit

bluca: Yes the socket unit itself has a WantedBy

jrybar-rh: True, I had it right before my eyes. On the first run on VM, something weird happened and the polkit helper service unit failed after boot for some reason, so I started to wonder... It went all good on second provision.

PR #505, opened 2024-10-10
Update Brazilian Portuguese translation

rffontenelle:
See https://l10n.gnome.org/vertimus/polkit/main/po/pt_BR/

### DIFF
--- a/po/pt_BR.po
+++ b/po/pt_BR.po
@@ -1,36 +1,37 @@
 # Brazilian Portuguese translation for polkit.
-# Copyright (C) 2018 polkit's COPYRIGHT HOLDER
+# Copyright (C) 2024 polkit's COPYRIGHT HOLDER
 # This file is distributed under the same license as the polkit package.
-# Rafael Fontenelle <rafaelff@gnome.org>, 2015, 2018.
+# Rafael Fontenelle <rafaelff@gnome.org>, 2015-2024.
+#
 msgid ""
 msgstr ""
 "Project-Id-Version: polkit master\n"
-"Report-Msgid-Bugs-To: https://gitlab.freedesktop.org/polkit/polkit/issues\n"
-"POT-Creation-Date: 2018-08-20 21:47+0000\n"
-"PO-Revision-Date: 2018-09-02 22:46-0200\n"
+"Report-Msgid-Bugs-To: https://github.com/polkit-org/polkit/issues\n"
+"POT-Creation-Date: 2024-03-19 15:27+0000\n"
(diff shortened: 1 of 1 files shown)


po/pt_BR.po
```diff
@@ -156,51 +157,58 @@ msgstr "%s: Sujeito não especificado\n"
...
 
 #. Translators: message shown when trying to run a program as another user.
 #. * Do not translate the $(program) or $(user) fragments - the former will
 #. * be expanded to the path of the program e.g. "/bin/bash" and the latter
 #. * to the user e.g. "John Doe (johndoe)" or "johndoe".
 #.
-#: ../src/programs/pkexec.c:800
-msgid "Authentication is needed to run `$(cmdline_short)' as user $(user.display)"
+#: src/programs/pkexec.c:832
+msgid ""
```
jrybar-rh, line 172: blank string followed by a non-blank string.

rffontenelle: I don't see a blank string. The difference is only that a line wrap was applied (by my gtranslator - my translation editor - or by Damned Lies). What before it was in the line 170 is now in 172 and 173, but it is still a valid msgid source message

jrybar-rh: What I see is replacement of `msgid "Authentication..."` with `msgid ""\n"Authentication"`. Yeah, it builds, no real impact. What I say is that it's unnecessary change (no other change in that english string).

rffontenelle: I'm sorry about that, but it wasn't a deliberate change. When gettext tools do line wrap, it applies to both msgid and msgstr. So if it applies line wrap in Nth character, it results in `msgid ""` alone followed by the rest of the text.

Let's me give you a particular example. I'm going to remove line wrap and then edit with Gtranslator.

Running `msgcat --no-wrap pt_BR.po > tmp && mv tmp pt_BR.po`, I get:

```gettext
#: src/programs/pkexec.c:832
msgid "Authentication is needed to run `$(cmdline_short)' as user $(user.display)"
msgstr "A autenticação é necessária para executar `$(cmdline_short)' como o usuário  $(user.display)"
```

Then editing with gtranslation (just added a space, removed, and saved), I get:

```gettext
#: src/programs/pkexec.c:832
msgid ""
"Authentication is needed to run `$(cmdline_short)' as user $(user.display)"
msgstr ""
"A autenticação é necessária para executar `$(cmdline_short)' como o usuário  "
"$(user.display)"

```

jrybar-rh: Interesting info. Thanks for your explanation =b

Issue #507, opened 2024-10-10
The LogLevel/LogTarget D-Bus properties can be set by unprivileged users


evverx:
**Describe the bug**
Unprivileged users can change the Log* D-Bus properties.

**To Reproduce**
```
$ busctl get-property org.freedesktop.PolicyKit1 /org/freedesktop/LogControl1 org.freedesktop.LogControl1 LogLevel
s "debug"
$ busctl set-property org.freedesktop.PolicyKit1 /org/freedesktop/LogControl1 org.freedesktop.LogControl1 LogLevel "s" info
$ busctl get-property org.freedesktop.PolicyKit1 /org/freedesktop/LogControl1 org.freedesktop.LogControl1 LogLevel
s "info"
```

**Expected behavior**
Unprivileged users shouldn't be able to change those properties. Here's what systemd-resolved does for example
```sh
$ busctl set-property org.freedesktop.resolve1 /org/freedesktop/LogControl1 org.freedesktop.LogControl1  LogLevel "s" info
Failed to set property LogLevel on interface org.freedesktop.LogControl1: Access denied
```
It's from https://github.com/polkit-org/polkit/commit/79d84d0dd0dd7d16c616555acf8cd14862d02baf

evverx:
The feature was reverted. I think it can be closed.

Issue #506, opened 2024-10-10
polkit crashes when the LogTarget D-Bus property is set


evverx:
**Describe the bug**
```sh
$ busctl set-property org.freedesktop.PolicyKit1 /org/freedesktop/LogControl1 org.freedesktop.LogControl1 LogTarget "s" ''
Failed to set property LogTarget on interface org.freedesktop.LogControl1: Remote peer disconnected
```
```
#0  0x00007f11e1a80724 in __pthread_kill_implementation () from target:/lib64/libc.so.6
#1  0x00007f11e1a27d0e in raise () from target:/lib64/libc.so.6
#2  0x00007f11e1a0f942 in abort () from target:/lib64/libc.so.6
#3  0x00007f11e24e414c in g_assertion_message[cold] () from target:/lib64/libglib-2.0.so.0
#4  0x00007f11e25538c7 in g_assertion_message_expr () from target:/lib64/libglib-2.0.so.0
#5  0x00007f11e270887d in invoke_set_property_in_idle_cb () from target:/lib64/libgio-2.0.so.0
#6  0x00007f11e2527add in g_idle_dispatch () from target:/lib64/libglib-2.0.so.0
#7  0x00007f11e252128c in g_main_context_dispatch_unlocked.lto_priv () from target:/lib64/libglib-2.0.so.0
#8  0x00007f11e25817f8 in g_main_context_iterate_unlocked.isra () from target:/lib64/libglib-2.0.so.0
#9  0x00007f11e2527377 in g_main_loop_run () from target:/lib64/libglib-2.0.so.0
#10 0x0000000000403a32 in main (argc=<optimized out>, argv=<optimized out>) at ../src/polkitbackend/polkitd.c:294
```
**To Reproduce**
```sh
busctl set-property org.freedesktop.PolicyKit1 /org/freedesktop/LogControl1 org.freedesktop.LogControl1 LogTarget "s" ''
```
It can also be triggered by just running dfuzzer (FYI @mrc0mmand)

**Expected behavior**
It shouldn't crash.

Looks like it's from 79d84d0dd0dd7d16c616555acf8cd14862d02baf

vmihalko:
[...]

PR #513, opened 2024-10-24
Revert "Implement LogControl1 protocol for dynamic log level changes"

jrybar-rh:
After a long discussion, the polkit team decided to revert the implementation of LogControl1 in polkit for the following reasons:
1. The authentication/authorization part of changing LogLevel needs to be implemented. Without it, any user can change log level via busctl (https://github.com/polkit-org/polkit/issues/507)
2. https://github.com/polkit-org/polkit/issues/506  

After all these issues are addressed, integration of the protocol can be later reconsidered.
LogControl1 is not widely used so far, therefore no major damage will be caused by postponing the integration of the protocol within polkit.

This reverts commit 79d84d0dd0dd7d16c616555acf8cd14862d02baf.

### DIFF
--- a/src/polkitbackend/polkitbackendauthority.c
+++ b/src/polkitbackend/polkitbackendauthority.c
@@ -502,7 +502,6 @@ polkit_backend_authority_revoke_temporary_authorization_by_id (PolkitBackendAuth
 typedef struct
 {
   guint authority_registration_id;
-  guint log_control_registration_id;
 
   GDBusNodeInfo *introspection_info;
 
@@ -526,9 +525,6 @@ server_free (Server *server)
 
   if (server->authority_registration_id > 0)
     g_dbus_connection_unregister_object (server->connection, server->authority_registration_id);
-
-  if (server->log_control_registration_id > 0)
-    g_dbus_connection_unregister_object (server->connection, server->log_control_registration_id);
 
   if (server->connection != NULL)
     g_object_unref (server->connection);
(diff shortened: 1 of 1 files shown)


bluca:
> LogControl1 is not widely used so far, therefore no major damage will be caused by postponing the integration of the protocol within polkit.

It actually is used - if you have verified yourself that there really are some bugs, why not simply fix them?

evverx (approved):
Thank you for reverting this!

As discussed in https://github.com/polkit-org/polkit/issues/506 I don't think this feature is particularly useful (even if the local unprivileged DoS and the incorrect authorization were fixed) so personally I think it's the best way to deal with it.

I also opened https://github.com/systemd/systemd/issues/34735 so as not to forget that the documentation can be improved in terms of how it all should actually be integrated into actual system daemons and maybe make it clear how systemd itself and its daemons actually work. Looks like it isn't clear enough.

jrybar-rh:
> > LogControl1 is not widely used so far, therefore no major damage will be caused by postponing the integration of the protocol within polkit.
> 
> It actually is used - if you have verified yourself that there really are some bugs, why not simply fix them?

We will fix it on separate branch unless you're interested in implementing the authorization part. I've tried to look around the debian codebase to see how other daemons implement authorization on LogControl1 requests over Dbus, but it seemed that polkit is the sole package to actually use that.  
But in this state, it must be reverted also in distros ASAP.

evverx:
@poettering can you please explain to @bluca what the "Privileged" part in
```
      @org.freedesktop.DBus.Property.EmitsChangedSignal("false")
      @org.freedesktop.systemd1.Privileged("true")
      readwrite s LogTarget = '...';
```
means?

jrybar-rh:
@evverx, can you please rephrase your last question? Thank you.

evverx:
@jrybar-rh sorry! I replied to the "There's no auth required" part. https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.LogControl1.html clearly says
```
      @org.freedesktop.DBus.Property.EmitsChangedSignal("false")
      @org.freedesktop.systemd1.Privileged("true")
      readwrite s LogTarget = '...';

Those two properties are writable, so they may be set by sufficiently privileged users
```
and https://github.com/polkit-org/polkit/issues/507 even shows how systemd-resolved works when unprivileged users try to set that property. It isn't that hard to take a look at the systemd codebase either. When things like that happen in the systemd repository @poettering usually helps to clear things up (and that's why I asked @poettering to do that). Either way https://github.com/systemd/systemd/issues/34735 should hopefully help eventually.

PR #520, opened 2024-11-04
Fix a couple of compilation warnings when building with clang

mrc0mmand:
There's still a bunch of warnings caused by `-Wdeprecated-declarations`, but these can be easily suppressed by `-Dc_args=-Wno-deprecated-declarations` for now.

This PR also tweaks the CI jobs to build with `-Werror`, so we should be able to catch these kind of issues a bit earlier in the future.

### DIFF
--- a/.github/workflows/ci.sh
+++ b/.github/workflows/ci.sh
@@ -3,8 +3,7 @@ set -eux
 set -o pipefail
 
 # TODO
-#   - enable --werror
-#       - currently there's a lot of warnings which need to be taken care of first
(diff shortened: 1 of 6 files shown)


evverx:
I think it would be great if it was possible to bring back clang in https://packit.dev/posts/openscanhub-prototype to be able to see all the clang warnings fixed in PRs (when `-Werror` isn't an option). As far as I can remember it was discussed somewhere but I think clang was turned off to reduce the number of false positives in projects using cleanup attributes.

Issue #515, opened 2024-10-29
dfuzzer should probably be run by the CI


evverx:
**Is your feature request related to a problem? Please describe.**

`polkit` is reachable via D-Bus so it would be great to poke its methods and properties to prevent things like https://github.com/polkit-org/polkit/issues/506 from happening again.

**Describe the solution you'd like**

It's possible to point https://github.com/dbus-fuzzer/dfuzzer to polkit to poke its D-Bus API (and https://github.com/polkit-org/polkit/issues/506 could have been easily found by just running dfuzzer). It's not sophisticated but it's enough to catch shallow issues. It can more or less detect crashes but Ii would be great to run polkit under ASan/UBSan to catch various memory corruptions like https://github.com/systemd/systemd/issues/23312 and https://github.com/systemd/systemd/issues/23097.

I set it up downstream but it obviously doesn't help to prevent issues in the first place when PRs are opened.

PR #521, opened 2024-11-04
Shutdown polkitd properly on SIGTERM

mrc0mmand:
While trying to break polkit using dfuzzer I couldn't make ASan/LSan
report _any_ memory leaks no matter how obvious they were. After a lot
of tinkering I managed to get the reports only when calling
`__lsan_do_recoverable_leak_check()` explicitly, which told me that the
shutdown `__lsan_do_leak_check() `is somehow getting skipped.

Turns out that polkitd does a "proper" shutdown on `SIGINT` (with which
ASan/LSan worked as expected), but not for `SIGTERM`. Fixing this, by
stopping the even loop even on SIGTERM, makes ASan/LSan happy (except
for the couple of memory leaks that were unfortunately hidden by this
behavior).

---

This PR together with #520 should be, for now, hopefully enough to let me introduce dfuzzer into polkit's test suite and address https://github.com/polkit-org/polkit/issues/515.

/cc @evverx

### DIFF
--- a/src/polkit/polkitsystembusname.c
+++ b/src/polkit/polkitsystembusname.c
@@ -566,6 +566,7 @@ polkit_system_bus_name_get_creds_sync (PolkitSystemBusName           *system_bus
     }
 
   g_variant_unref (result);
+  g_variant_iter_free (iter);
 
   if (out_uid)
     *out_uid = uid;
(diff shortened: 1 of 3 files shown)


evverx:
I think it should also help to build coverage reports.

Issue #517, opened 2024-10-31
polkit and Coverity Scan


evverx:
**Is your feature request related to a problem? Please describe.**
I went to https://scan.coverity.com/ and found https://scan.coverity.com/projects/polkit there. Looks like it hasn't been updated since 2014.

**Describe the solution you'd like**
 It would be great to send polkit to Coverity Scan automatically by analogy with https://github.com/bus1/dbus-broker/issues/316.

**Describe alternatives you've considered**
I can send it there manually by analogy with that I do with dbus-broker but it's not ideal.

pwithnall:
I’m currently the only admin for that Coverity account. I’m happy to make the current maintainers of polkit also be admins there. I don’t have time to maintain or run the Coverity scans any more. Is there a canonical list of the current maintainers somewhere?

evverx:
I think in terms of setting up a GitHub action sending data to Coverity Scan automatically it should be enough to add @mrc0mmand there to test the integration like https://github.com/bus1/dbus-broker/pull/363. I don't know if @mrc0mmand has access to the repository secrets here on GitHub but I don't think it should be a problem to pass the coverity token to the maintainers with that kind of access to the repository.

pwithnall:
@jrybar-rh, who are the current maintainers of polkit? From recent commit history it looks like just you have merge rights. I’d love to give permissions for Coverity to someone, but I want to double check I’m giving it to the right people!

mrc0mmand:
@pwithnall I think you can give the permissions to @jrybar-rh for now and he can then extend this to other people when needed.

pwithnall:
I’ve invited @jrybar-rh to Coverity using their redhat.com address

jrybar-rh:
Thank you, @pwithnall.

mrc0mmand:
@pwithnall do you want stay as a member or be removed completely from the Coverity project?

pwithnall:
Please remove me completely :)

mrc0mmand:
> Please remove me completely :)

Done. Thanks a lot for the access to the project!

evverx:
> BTW polkit is tested in OSH for Fedora, just sayin'

[...]

PR #523, opened 2024-11-06
ci: do nightly polkit builds & submit them to Coverity

mrc0mmand:
_Note: I created this PR's branch intentionally directly in the upstream repo, so it has access to the repo secrets with the Coverity token, as PRs from forks can't do that._

Let's reintroduce regular Coverity builds. Since there's a pretty strict rate limit [0], do one nightly build each day, and upload it to Coverity for analysis. The results can be then found in the project dashboard [1].

[0] https://scan.coverity.com/faq#frequency
[1] https://scan.coverity.com/projects/polkit?tab=overview

Resolves: #517

### DIFF
--- a/.github/workflows/coverity.sh
+++ b/.github/workflows/coverity.sh
@@ -1,0 +1,62 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-2.0-or-later
+
+set -eux
+
+COVERITY_SCAN_TOOL_BASE="/tmp/coverity-scan-analysis"
+COVERITY_SCAN_PROJECT_NAME="polkit"
+
+coverity_install_script() {
+    local platform tool_url tool_archive
+
+    platform=$(uname)
+    tool_url="https://scan.coverity.com/download/${platform}"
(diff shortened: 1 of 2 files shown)


.github/workflows/coverity.sh
```diff
@@ -0,0 +1,62 @@
...
+
+run_coverity() {
+    local results_dir tool_dir results_archive sha response status_code
+
+    results_dir="cov-int"
+    tool_dir=$(find "$COVERITY_SCAN_TOOL_BASE" -type d -name 'cov-analysis*')
+    results_archive="analysis-results.tgz"
+    sha=$(git rev-parse --short HEAD)
+
+    meson setup build -Dtests=true -Dexamples=true
```
evverx, line 35: I think it would be great to turn on as many features as possible to cover things like https://github.com/polkit-org/polkit/pull/486 as well. On the other hand the default configuration is more important probably. I'm not sure.

mrc0mmand: Good point. I went through the available options and thought I enabled all that could be relevant for Coverity, but I apparently missed some :)

PR #526, opened 2024-11-07
test: run dfuzzer against Polkit's D-Bus interface

mrc0mmand:
Let's run dfuzzer to fuzz polkit's own D-Bus interface to catch issues
like [0][1][2] early.

The test implementation is a _bit_ convoluted, caused by the fact that
we have to rebuild polkit with Address and Undefined sanitizers to make
dfuzzer considerably more effective. This also involves running the just
built polkit in a lightweight container so we don't mess with the
already installed polkit (which could affect tests running after this
one that don't expect sanitized polkit). All this extra code can be
potentially dropped once we're able to do multiple Packit builds in one
PR (currently in an RFC stage).

Resolves: https://github.com/polkit-org/polkit/issues/515

[0] https://github.com/polkit-org/polkit/issues/506
[1] https://github.com/polkit-org/polkit/commit/b709b693346712bf8e8d8e8c77f71b7918fa367d
[2] https://github.com/polkit-org/polkit/commit/5cce296a8d60a3395c1a372377bdf8f01ed93872

### DIFF
--- a/test/integration/dfuzzer/main.fmf
+++ b/test/integration/dfuzzer/main.fmf
@@ -1,0 +1,30 @@
+summary: Run dfuzzer against polkit's D-Bus interface
+test: ./test.sh
+require:
+    - clang
+    - compiler-rt
+    - coreutils
+    - dbus
+    - dbus-devel
+    - dfuzzer
+    - expat-devel
+    - gcc-c++
+    - gdb
+    - gettext-devel
+    - git
+    - glib2-devel
+    - glibc-devel
+    - gobject-introspection-devel
+    - gtk-doc
+    - libasan
+    - libubsan
+    - llvm
+    - meson
+    - pam-devel
+    - pkgconfig(duktape)
+    - python3-dbusmock
+    - systemd
+    - systemd-container
+    - systemd-devel
+    - util-linux
+duration: 30m
--- a/test/integration/dfuzzer/test.sh
+++ b/test/integration/dfuzzer/test.sh
@@ -1,0 +1,176 @@
+#!/bin/bash
+# vi: set sw=4 ts=4 et tw=110:
+# shellcheck disable=SC2016
+#
+# FIXME: this test can be _drastically_ simplified once we can run a dedicated sanitizer job, see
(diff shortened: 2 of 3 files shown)


test/integration/dfuzzer/main.fmf
```diff
@@ -0,0 +1,31 @@
+summary: Run dbus-broker under sanitizers
```
evverx, line 1:
```suggestion
summary: Run polkit under sanitizers
```

test/integration/dfuzzer/main.fmf
```diff
@@ -0,0 +1,31 @@
...
+    - git
+    - glib2-devel
+    - glibc-devel
+    - gobject-introspection-devel
+    - gtk-doc
+    - libasan
+    - libubsan
+    - llvm
+    - meson
+    - meson
```
evverx, line 23:
```suggestion
```

test/integration/dfuzzer/test.sh
```diff
@@ -0,0 +1,173 @@
...
+# can't just check ldd's output. Another option is using nm/objdump to check for ASan-specific functions, but
+# that's also error prone. Instead, let's call each binary with ASan's "help" option, which produces output
+# only if the target binary is built with (hopefully working) ASan.
+container_run bash -xec 'ASAN_OPTIONS=help=1 /proc/$(systemctl show -P MainPID polkit.service)/exe -h 2>&1 >/dev/null | grep -q AddressSanitizer'
+container_run systemctl show -p Environment polkit.service | grep -q ASAN_OPTIONS
+
+# Now we should have a container ready for our shenanigans
+
+# Fuzz polkit's own interface
+run_and_check dfuzzer -v -n org.freedesktop.PolicyKit1
```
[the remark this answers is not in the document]
mrc0mmand: Ah, indeed. The coredump check is executed always, but the backtrace check is not. Will fix, thanks!

test/integration/dfuzzer/main.fmf
mrc0mmand: That's a backup meson, just in case the first one doesn't work :)

PR #527, opened 2024-11-13
ci: run Packit on commit events as well

mrc0mmand:
Let's do a Packit build on "commit" events as well, which automagically enables Packit's OpenScanHub integration [0].

Since OSH runs only on x86_64 Fedora Rawhide builds, let's limit the scope for the "commit" event adequately; we still do the usual builds across all supported Fedora releases (and architectures) in PRs.

The OSH integration is still in its experimental stage, but there's a plan to add SARIF support and integrate it into GitHub's UI (see [1]), which should make it quite useful.

[0] https://packit.dev/posts/openscanhub-prototype
[1] https://github.com/packit/dashboard/issues/441

---

Note: this won't have any effect until it's merged, as OSH needs at least one build from the default (main) branch to have something to compare against.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -16,6 +16,12 @@ downstream_package_name: polkit
 
 jobs:
 - job: copr_build
+  trigger: commit
+  preserve_project: true
+  targets:
+  - fedora-rawhide-x86_64
+
+- job: copr_build
   trigger: pull_request
   targets:
   - fedora-all-aarch64


siteshwar:
An example scan can be seen at https://openscanhub.fedoraproject.org/task/23951/.

evverx:
Nice!

(I still think it would be great if it was possible to (optionally) bring clang back :-))

lbarcziova:
@siteshwar I assume this is what https://github.com/packit/packit/issues/2391 solve, right? We may prioritise that.

siteshwar:
> @siteshwar I assume this is what [packit/packit#2391](https://github.com/packit/packit/issues/2391) would solve, right? We may prioritise that.

I believe that issue is too generic and unless you are planning to design how the configurations may look in advance, each options may have to be taken individually based on user requests.

PR #529, opened 2024-11-20
A couple of fixes for issues discovered by GCC analyzer

mrc0mmand:
This addresses three issues (and one potential one) found by GCC analyzer after my adventures with OpenScanHub. See the respective commits for more details.

### DIFF
--- a/src/polkit/polkitpermission.c
+++ b/src/polkit/polkitpermission.c
@@ -482,6 +482,7 @@ changed_check_cb (GObject       *source_object,
       /* this really should never fail (since we are not passing any
        * details) so log to stderr if it happens
        */
+      g_assert (error != NULL);
       g_warning ("Error checking authorization for action id %s: %s",
                  permission->action_id,
                  error->message);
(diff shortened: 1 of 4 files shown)


siteshwar:
Thanks! This should be enabled in Packit configurations too. Please add this line to Packit configurations:

```
csmock_args: --gcc-add-flag=-flto=auto
```

siteshwar:
OpenScanHub task shows 4 fixed warnings after LTO was enabled: https://openscanhub.fedoraproject.org/task/25069/log/fixed.html

So, it seems to work as expected. Thanks!

.packit.yaml
```diff
@@ -13,6 +13,8 @@ files_to_sync:
 upstream_package_name: polkit
 # downstream (Fedora) RPM package name
 downstream_package_name: polkit
+# arguments for OpenScanHub
```
siteshwar, line 16: You may want to mention why is it enable here. For example, it only works at smaller scale as of now and can not be enabled for larger projects, but improves quality of GCC analyzer reports.

mrc0mmand: I had this originally only in the commit message, but I agree that having it in the configuration file should be better for future traceability. Updated.

siteshwar (approved):

PR #531, opened 2024-12-03
fix: need goto out after g_set_error

KT-lcz:
fix: need goto out after g_set_error

goto out needs to be executed after g_set_error , otherwise the assertion will be triggered in g_assert (process != NULL);


jrybar-rh:
Thanks, @KT-lcz.

PR #532, opened 2024-12-15
Two small fixes

bluca:


bluca:
The property is defined with a min var of 0, so there's a warning printed as the commit id shows:

https://github.com/polkit-org/polkit/blob/c1369cd4ffb975e177035392837ff91373774407/src/polkit/polkitunixprocess.c#L413

See for example here where it's passed directly to g_variant_new_uint32:

https://github.com/polkit-org/polkit/blob/main/src/polkit/polkitsubject.c#L313

I've fixed that one case, everything else checks `<= 0` for validity

bluca:
Can we merge this, then? Thanks

jrybar-rh:
> Can we merge this, then? Thanks

We sure can and we probably will, on today's meeting :)

PR #533, opened 2024-12-16
auth_keep: do not ask for reauth if new process shares same UID/parent/cgroup/tty

### LINKED ISSUE #472: Remember authentication for a set period of time
**Is your feature request related to a problem? Please describe.**
When running security-sensitive actions such as managing services via `systemctl`, Polkit asks me to authenticate, which is not a bad thing, but I will often run many such actions in a row, requiring me to retype the password over and over (remember that a secure password should consist of multiple words).

The well-known utility `sudo` also deals with a security-sensitive action (running as root), but it remembers the password for a configurable amount of time (15 minutes by default), for a given session (a different session will be asked for the password even in that period; not sure how a session is determined by `sudo`, but the important part is that a single shell process is counted as a single session). Ironically, this makes me often run `systemctl` as `sudo systemctl`, since root is not prompted for a password by Polkit (I assume this is the upstream configuration, though I use the Debian package). This is a pretty stupid workaround, but seems like many do it anyways just because of this usability issue.

This is even more important with the addition of a `run0` utility to the systemd suite, which is meant to be a cleaner, safer alternative to `sudo` based on the existing underlying systemd security system, including Polkit. While it does a quite good job, `run0` will now ask you for the password *every time you run something as root*, because that's what Polkit does. Since `sudo` offered the capacity to remember the password, this is a roadblock to `run0` adoption, at least in my case, and I imagine many others.

**Describe the solution you'd like**
Add a way to configure a period of time during which Polkit will remember the entered password for this session, default to 15 minutes like `sudo`. A session in CLI should encompass still using the same shell process, in whatever way this is done, in GUI it should mean using the same desktop session. This could still require confirmation, as described in #463.

With this, there are a few issues with the CLI authentication prompt itself:

- If there's more than one user that can do the action, it will first ask for the user to authenticate as, which is inconvenient, but I'm not sure how it should be resolved (the GUI prompt has the advantage of being able to just leave the selection as default; maybe a keyboard shortcut could be used to explicitly change the user from the default?).
- Using the utility `run0`, which only purpose is to run things as other users (which is security sensitive) probably should be considered enough confirmation by itself (entering `y` and enter every time would still be pointless, though better than the current situation). This seems problematic to do properly with Polkit, and it has no analogue in the GUI world.
- The GUI prompt is launched even in CLI contexts, when the initial shell is launched from a GUI terminal.

Those would probably better be explored in separate issues, however.

- The status quo: password authentication every time (as explained previously: assuming a secure password - a PITA).
- Biometric authentication? I don't know if/how that works on Linux/Polkit/PAM/anything, I don't have the hardware for it (most people don't) and I don't even know if I trust that technology fundamentally.
- Authorization solely based on being a certain user/in a certain group, which is already possible with Polkit, or confirmation-only authorization, not yet possible - see #463, where I left a comment.

**Additional context**
I think the utility of admin passwords should really be reconsidered, as described by my comment in #463. Still, it would be good to see this as an option for people that prefer having them, or just simply if having confirmation-only authorization be the default is not currently (or ever) considered desirable.

@poettering Curious about your thoughts, since you seemed excited for `run0` to take over the world ;) and probably have a better idea about how this should be done to benefit systemd tooling. Seems risky to ping, but pinging the entire systemd project would be worse, especially considering my previous issue...

bluca:
sudo keeps a record of authenticated processes via either the controlling TTY (default) or the parent process id.

Implement the same caching behaviour, but stricter: if a process is authenticated for auth_keep, do not expunge it when it exits if it was tracked via PID FD (to make it safe against reuse attacks).

Then, if another process comes along, skip re-auth and allow it if it shared the same UID, parent process id, cgroup id and controlling terminal (and all processes are newer than the controlling terminal ctime). PID FDs must be used all the way through, otherwise there's no caching.

This is much stricter than sudo, as all conditions must be met. But it still allows to fulfill the main use case, which is to run multiple commands on the same terminal without being asked for the password again and again.

Unlike sudo, we also do not refresh the countdown on each use.

Fixes #472

This allows running `run0` or `systemctl` and friends multiple times from the same terminal without having to re-enter the password multiple times, if within the default temporary auth_keep time window (5 minutes)

### DIFF
--- a/src/polkit/polkitunixprocess.c
+++ b/src/polkit/polkitunixprocess.c
@@ -19,6 +19,13 @@
  * Author: David Zeuthen <davidz@redhat.com>
  */
 
+#ifdef __linux__
+#include <fcntl.h>
+#include <sys/ioctl.h>
+#include <sys/stat.h>
+#include <sys/sysmacros.h>
+#include <sys/vfs.h>
+#endif
 #include <sys/types.h>
 #ifdef HAVE_FREEBSD
 #include <sys/param.h>
@@ -151,8 +158,11 @@ struct _PolkitUnixProcess
 
   gint pid;
   guint64 start_time;
+  guint64 cgroupid;
   gint uid;
   gint pidfd;
+  gint ppidfd;
+  guint ctty;
   gboolean pidfd_is_safe;
   GArray *gids;
 };
@@ -166,10 +176,14 @@ enum
 {
   PROP_0,
   PROP_PID,
+  PROP_PPID,
   PROP_START_TIME,
+  PROP_CGROUPID,
   PROP_UID,
   PROP_PIDFD,
   PROP_PIDFD_IS_SAFE,
+  PROP_PPIDFD,
+  PROP_CTTY,
   PROP_GIDS,
 };
 
@@ -177,6 +191,18 @@ static void subject_iface_init (PolkitSubjectIface *subject_iface);
 
 static guint64 get_start_time_for_pid (gint    pid,
                                        GError **error);
+
+static gint
+get_ppidfd_for_pidfd (gint     pidfd,
+                      GError **error);
+
+static guint
+get_ctty_number_for_pidfd (gint     pidfd,
+                           GError **error);
+
+static guint64
+get_cgroupid_for_pidfd (gint     pidfd,
+                        GError **error);
 
 #if defined(HAVE_FREEBSD) || defined(HAVE_NETBSD) || defined(HAVE_OPENBSD)
 static gboolean get_kinfo_proc (gint pid,
@@ -196,6 +222,7 @@ polkit_unix_process_init (PolkitUnixProcess *unix_process)
 {
   unix_process->uid = -1;
   unix_process->pidfd = -1;
+  unix_process->ppidfd = -1;
 }
 
 static void
@@ -212,6 +239,10 @@ polkit_unix_process_get_property (GObject    *object,
       g_value_set_int (value, polkit_unix_process_get_pid (unix_process));
       break;
 
+    case PROP_PPID:
+      g_value_set_int (value, polkit_unix_process_get_ppid (unix_process));
+      break;
+
     case PROP_UID:
       g_value_set_int (value, unix_process->uid);
       break;
@@ -224,12 +255,24 @@ polkit_unix_process_get_property (GObject    *object,
       g_value_set_int (value, unix_process->pidfd);
       break;
 
+    case PROP_PPIDFD:
+      g_value_set_int (value, unix_process->ppidfd);
+      break;
+
+    case PROP_CTTY:
+      g_value_set_uint (value, unix_process->ctty);
+      break;
+
     case PROP_PIDFD_IS_SAFE:
       g_value_set_boolean (value, unix_process->pidfd_is_safe);
       break;
 
     case PROP_START_TIME:
       g_value_set_uint64 (value, unix_process->start_time);
+      break;
+
+    case PROP_CGROUPID:
+      g_value_set_uint64 (value, unix_process->cgroupid);
       break;
 
     default:
@@ -275,8 +318,8 @@ polkit_unix_process_set_property (GObject      *object,
 }
 
 static gint
-polkit_unix_process_get_pid_from_pidfd (PolkitUnixProcess  *process,
-                                        GError            **error)
+polkit_unix_process_get_pid_from_pidfd (gint   pidfd,
+                                        GError **error)
 {
   gint result;
   gchar *contents;
@@ -284,15 +327,14 @@ polkit_unix_process_get_pid_from_pidfd (PolkitUnixProcess  *process,
   gchar filename[64];
   guint n;
 
-  g_return_val_if_fail (POLKIT_IS_UNIX_PROCESS (process), -1);
+  g_return_val_if_fail (pidfd >= 0, -1);
   g_return_val_if_fail (error == NULL || *error == NULL, -1);
-  g_return_val_if_fail (process->pidfd >= 0, -1);
 
   result = -1;
   lines = NULL;
   contents = NULL;
 
-  g_snprintf (filename, sizeof filename, "/proc/self/fdinfo/%d", process->pidfd);
+  g_snprintf (filename, sizeof filename, "/proc/self/fdinfo/%d", pidfd);
   if (!g_file_get_contents (filename,
                             &contents,
                             NULL,
@@ -351,6 +393,31 @@ polkit_unix_process_constructed (GObject *object)
           process->pid = 0;
         }
     }
+
+    if (process->pidfd >= 0)
+      {
+        GError *error;
+        error = NULL;
+
+        process->ppidfd = get_ppidfd_for_pidfd (process->pidfd, &error);
+        if (error != NULL)
+          {
+            g_error_free (error);
+            error = NULL;
+            process->ppidfd = -1;
+          }
+
+        process->cgroupid = get_cgroupid_for_pidfd (process->pidfd, &error);
+        if (error != NULL)
+          {
+            g_error_free (error);
+            error = NULL;
+          }
+
+        process->ctty = get_ctty_number_for_pidfd (process->pidfd, &error);
+        if (error != NULL)
+          g_error_free (error);
+      }
 #endif /* HAVE_PIDFD_OPEN */
 
   if (process->start_time == 0)
@@ -381,6 +448,12 @@ polkit_unix_process_finalize (GObject *object)
     {
       close (process->pidfd);
       process->pidfd = -1;
+    }
+
+  if (process->ppidfd >= 0)
+    {
+      close (process->ppidfd);
+      process->ppidfd = -1;
     }
 
   if (process->gids)
@@ -460,6 +533,24 @@ polkit_unix_process_class_init (PolkitUnixProcessClass *klass)
                                                         G_PARAM_STATIC_NICK));
 
   /**
+   * PolkitUnixProcess:cgroupid:
+   *
+   * The start time of the process.
+   */
+  g_object_class_install_property (gobject_class,
+                                   PROP_CGROUPID,
+                                   g_param_spec_uint64 ("cgroupid",
+                                                        "ControlGroup ID",
+                                                        "The ID of the Control Group of the process",
+                                                        0,
+                                                        G_MAXUINT64,
+                                                        0,
+                                                        G_PARAM_READABLE |
+                                                        G_PARAM_STATIC_NAME |
+                                                        G_PARAM_STATIC_BLURB |
+                                                        G_PARAM_STATIC_NICK));
+
+  /**
    * PolkitUnixProcess:pidfd:
    *
    * The UNIX process id file descriptor.
@@ -477,6 +568,42 @@ polkit_unix_process_class_init (PolkitUnixProcessClass *klass)
                                                      G_PARAM_STATIC_NAME |
                                                      G_PARAM_STATIC_BLURB |
                                                      G_PARAM_STATIC_NICK));
+
+  /**
+   * PolkitUnixProcess:ppidfd:
+   *
+   * The UNIX process' parent id file descriptor.
+   */
+  g_object_class_install_property (gobject_class,
+                                   PROP_PPIDFD,
+                                   g_param_spec_int ("ppidfd",
+                                                     "Process' parent ID FD",
+                                                     "The UNIX process' parent ID file descriptor",
+                                                     -1,
+                                                     G_MAXINT,
+                                                     -1,
+                                                     G_PARAM_READABLE |
+                                                     G_PARAM_STATIC_NAME |
+                                                     G_PARAM_STATIC_BLURB |
+                                                     G_PARAM_STATIC_NICK));
+
+  /**
+   * PolkitUnixProcess:ctty:
+   *
+   * The UNIX process controlling TTY.
+   */
+  g_object_class_install_property (gobject_class,
+                                   PROP_CTTY,
+                                   g_param_spec_uint ("ctty",
+                                                      "Process controlling TTY",
+                                                      "The UNIX process controlling TTY",
+                                                      0,
+                                                      G_MAXINT,
+                                                      0,
+                                                      G_PARAM_READABLE |
+                                                      G_PARAM_STATIC_NAME |
+                                                      G_PARAM_STATIC_BLURB |
+                                                      G_PARAM_STATIC_NICK));
 
   /**
    * PolkitUnixProcess:pidfd_is_safe:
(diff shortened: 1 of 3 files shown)


src/polkit/polkitunixprocess.c
```diff
@@ -151,8 +156,11 @@ struct _PolkitUnixProcess
 
   gint pid;
   guint64 start_time;
+  guint64 cgroupid;
   gint uid;
   gint pidfd;
+  gint ppidfd;
```
poettering, line 162: i don't think looking at the ppid is a good idea, because that means subshells (some of which are somewhat hidden in shells) would not qualify anymore.

instead: check the posix session id (i.e. the one managed by setsid()/getsid(), not the audit session id, nor the logind session id, which are three distinct things). the posix session id should be unique for each shell, and is necessary for correct shell sesson mgmt (it means that if the shell goes away all forked of processes get SIGHUP). 

the session leader has pid=sid btw, and if it exits the session becomes orphaned. (thus the sid will exist for longer, until all processes with it actually exited, but now without a leader). but that's good enough, it allows tracking sids via pidfds for this purpose, and thus get a proper pidfd inode id for the sid, that is safe regarding recycling

src/polkit/polkitunixprocess.c
```diff
@@ -1131,6 +1361,437 @@ get_start_time_for_pid (pid_t    pid,
...
+    {
+      g_set_error (error,
+                   POLKIT_ERROR,
+                   POLKIT_ERROR_FAILED,
+                   "Error allocating memory for cgroup path");
+      goto out;
+    }
+
+  /* We know in advance the size of the field, cgroupid is a uint64 */
+  fhp = malloc (offsetof(struct file_handle, f_handle) + sizeof(guint64));
```
poettering, line 1440: use g_malloc()? g_malloc() aborts on oom, hence no need to handle the oom case then

src/polkit/polkitunixprocess.c
```diff
@@ -1131,6 +1361,437 @@ get_start_time_for_pid (pid_t    pid,
...
+                   POLKIT_ERROR_FAILED,
+                   "Error parsing file %s",
+                   filename);
+      goto out;
+    }
+
+  contents[length - 1] = '\0'; /* remove the newline */
+
+  cgroup_path = g_strdup_printf ("/sys/fs/cgroup%s", p);
+  if (!cgroup_path)
```
poettering, line 1430: glib aborts on oom anyway, you don't need to check for oom hence. not here not anywhere else

src/polkit/polkitunixprocess.c
```diff
@@ -1131,6 +1361,437 @@ get_start_time_for_pid (pid_t    pid,
...
+  cgroupid = 0;
+  fhp = NULL;
+
+  g_return_val_if_fail (pidfd >= 0, 0);
+
+  pid = polkit_unix_process_get_pid_from_pidfd (pidfd, error);
+  if (pid <= 0)
+    return 0;
+
+  filename = g_strdup_printf ("/proc/%d/cgroup", pid);
```
poettering, line 1392: this is ugly and racy. maybe just use the new ioctl to get the cgroup id directly from the pid?

src/polkit/polkitunixprocess.c
```diff
@@ -1131,6 +1361,437 @@ get_start_time_for_pid (pid_t    pid,
...
+
+  /*
+   * As per the manpage, the minor device number is contained in the
+   * combination of bits 31 to 20 and 7 to 0, the major device number is in
+   * bits 15 to 8.
+   */
+  major = (tty_nr & 0xff00) >> 8;
+  minor = ((tty_nr & 0xfff00000) >> 12) | (tty_nr & 0xff);
+
+  if (major < 136 || major > 143)
```
poettering, line 1606: uh, this is ugly. why do this at all? if something shows up as ctty for a process then it's a tty, no need to validate that. just trust it?

if you want to properly check if something is a pty on linux, better look if backing fs is devpts, and if so check how it reacts to TIOCGPKT. See how terminal_is_pty_fd() does it in the systemd codebase.

but again: why restrict this to pty here? any tty should be fine.

and also: don't bother with ttys here at all, use getsid()/setsid() if anthing.

src/polkit/polkitunixprocess.c
poettering: because cgroups can be removed/recreated, and are different beasts then, as per cgroup id. if you go directly from pid to cgroupid you are not vulnerable to races around that hence.

src/polkit/polkitunixprocess.c
bluca: > uh, this is ugly. why do this at all? if something shows up as ctty for a process then it's a tty, no need to validate that. just trust it?

This crosses a security boundary, ie we are inspecting metadata of a foreign, untrusted process. So I thought it was a good idea to dot the Is and cross the Ts

src/polkit/polkitunixprocess.c
bluca: > i don't think looking at the ppid is a good idea, because that means subshells (some of which are somewhat hidden in shells) would not qualify anymore.

That was intentional: right now auth is checked on every single invocation. This PR relaxes this somewhat, by allowing multiple commands in the same shell to be done without checking auth again. Letting scripts running in the same shell also run without auth is a step too far for now. We can think about it again later, but for now I prefer to keep it very explicit, so that if you do:

```
$ pkexec foo
$ pkexec bar
```

You only get auth'ed once, but if you do:

```
$ pkexec foo
$ ./bar.sh
```

and bar.sh itself calls pkexec, auth is checked again. This was it's very very explicit, or at least it tries to be. We can consider further relaxing it more down the line. There are other things to consider in the future too, to match sudo's behaviour: whether to increase the default window to 15m, and whether to "refresh" the token on each invocation. But for now let's keep it nice and tight.

> the posix session id should be unique for each shell

- it _should_ be, but it's not mandatory, it depends on the shell implementation itself
- it is 32bit so it can wrap around, and the only way to make that safe is to use the super-new pidfd filesystem (introduced this year) which is 64bit only and requires a disabled-by-default kconfig to be enabled

so it's riskier, and cuts off basically all existing Linux systems, and might be usable in a year or two, so it's too restrictive, so for now I'll keep checking the tty

src/polkit/polkitunixprocess.c
poettering: don't second guess the kernel. 

i don't get you philosophy on this. one one hand you don't trust the kernel enough to make only ttys cttys, and second guess it hence. on the other hand you yolo around races...

src/polkit/polkitunixprocess.c
```diff
@@ -1131,6 +1362,445 @@ get_start_time_for_pid (pid_t    pid,
...
+
+  /* Now do some sanity checks to ensure we are really looking at a TTY. */
+
+  /*
+   * As per the manpage, the minor device number is contained in the
+   * combination of bits 31 to 20 and 7 to 0, the major device number is in
+   * bits 15 to 8.
+   */
+  major = (tty_nr & 0xff00) >> 8;
+  minor = ((tty_nr & 0xfff00000) >> 12) | (tty_nr & 0xff);
```
poettering, line 1586: sorry, this is nonsense.

src/polkit/polkitunixprocess.c
```diff
@@ -1131,6 +1362,445 @@ get_start_time_for_pid (pid_t    pid,
...
+
+  tty_name = g_strdup_printf ("/dev/pts/%lu", minor);
+
+  tty_fd = open (tty_name, O_RDONLY|O_CLOEXEC);
```
poettering, line 1590: oh, don't open random ttys like this. ttys are magic shit. at the very least you want O_NOCTTY, but note that opening ttys can be blocking too, you shouldn't just do that. use O_PATH, if you just care about th einode.

but frankly, this all is unnecessary. you can trust the kernel on only allowing ttys to become cttys, this validation is nonsensical.

also note that ptys are getting reused, hence reopening them iva /dev/pts is a mistake.

src/polkit/polkitunixprocess.c
poettering: there are macros for this, btw, minor() and major()

src/polkit/polkitunixprocess.c
bluca: Already removed, you might need to refresh the tab if GH still shows this

src/polkit/polkitunixprocess.c
```diff
@@ -1131,6 +1362,445 @@ get_start_time_for_pid (pid_t    pid,
...
+      g_set_error (error,
+                   POLKIT_ERROR,
+                   POLKIT_ERROR_FAILED,
+                   "Error getting TTY filesystem information for process %d: %s",
+                   pid,
+                   g_strerror (errno));
+      goto out;
+    }
+
+  if (stfs.f_type != 0x1cd1) /* 0x1cd1 is the magic number for devpts */
```
poettering, line 1645: you opened the device via /dev/pts/, hence this check is unnecessary, you know it's below /dev/pts...

poettering:
you shouldn't do those pty shenanigans, that helps noone, it just breaks the concept on non-pty ttys, such as the VT, serial, hvc, and so on.

and use the sid anyway, not the other stuff.

note that /proc/$PID/stat is a linux thing. as i understand you care a lot about solaris, so why not check the sid instead, that's posix after all...

src/polkit/polkitunixprocess.c
bluca: You told me to do this:

> if you want to properly check if something is a pty on linux, better look if **backing fs is devpts**, and if so check how it reacts to TIOCGPKT

https://github.com/polkit-org/polkit/pull/533#discussion_r1890338449

src/polkit/polkitunixprocess.c
poettering: > it should be, but it's not mandatory, it depends on the shell implementation itself

no it does not? the sid is always initialized from the pid of the caller. you don't get to choose your sid?

ptys are tightly packed (i.e. you always get lowest free pty), and aggressively recycled. otoh, sids are just like pids, i.e. allocated form a counter basically.

in other words, there is no rational rason why you think checking ptys is fine but checking sids is not.

sids change between user sessions on the same ttys, but tty numbers are super stable, hence not suitable for authentication that shall be context bound.

src/polkit/polkitunixprocess.c
poettering: again, you should not parse major/minor numbes at all, but if you do, use major()/minor() macros for this, instead of hand grown stuff.

src/polkit/polkitunixprocess.c
poettering: if you have an open fd, sure. but if you open the fd anew, and you already went through the path to get it then why check this again?

bluca:
> you shouldn't do those pty shenanigans, that helps noone, it just breaks the concept on non-pty ttys, such as the VT, serial, hvc, and so on.

Serial/hvc do not work anyway, as they are not local.

> and use the sid anyway, not the other stuff.

That's not enough, as it can wrap, it's a 32bit number

src/polkit/polkitunixprocess.c
bluca: > no it does not? the sid is always initialized from the pid of the caller. you don't get to choose your sid?

But that means, for all intents and purposes, sid == pid, including the fact that it wraps around and doesn't require privileges to get a new one, and hit a collision.

> sids change between user sessions on the same ttys, but tty numbers are super stable, hence not suitable for authentication that shall be context bound.

sudo uses ttys to cache

poettering:
> Serial/hvc do not work anyway, as they are not local.

This seems to be a pretty annoying pointless restriction. Whether a seat is available or not shouldn't matter for authentication. It should be fine to restric scope to the sid of a session (i.e. to one specific ssh session for example)

poettering:
> That's not enough, as it can wrap, it's a 32bit number

so is the tty, but with much worse properties. you cannot use small range set as excuse for using tty over sid, because the sid range is much larger and not tightly packed near the bottom.

bluca:
> > Serial/hvc do not work anyway, as they are not local.
> 
> This seems to be a pretty annoying pointless restriction. Whether a seat is available or not shouldn't matter for authentication. It should be fine to restric scope to the sid of a session (i.e. to one specific ssh session for example)

That may or may not be, but it's how it works now and how it always did, and I will not change it here, as it's out of scope

bluca:
> > That's not enough, as it can wrap, it's a 32bit number
> 
> so is the tty, but with much worse properties. you cannot use small range set as excuse for using tty over sid, because the sid range is much larger and not tightly packed near the bottom.

There is one crucial difference, and it's likely the reason sudo uses it: you can check the mtime of the tty, and ensure it is older than both processes, which is what I am doing here, and what sudo also does.

poettering:
uh? mtime of the tty? that's updated on every write() to the tty, on every command you invoke. we use that in logind for example to do idle detection of tty sections.

mtime is under full user control, should not be used for security decisions whatsoever. anyone can do `touch $(tty)`

bluca:
Check the diff, meant ctime, typing from phone

nyabinary:
bump?

Henry-Hiles:
Any blockers / why hasn't this been merged?

jrybar-rh:
> Any blockers / why hasn't this been merged?

Because it hasn't been reviewed yet?
It's **on the list**.

Henry-Hiles:
> > Any blockers / why hasn't this been merged?
> 
> Because it hasn't been reviewed yet? It's **on the list**.

Sorry, what is the list? And I was just wondering since it hasn't been reviewed for a few months, and it's quite important for stuff like `run0`.

jrybar-rh:
> > > Any blockers / why hasn't this been merged?
> > 
> > 
> > Because it hasn't been reviewed yet? It's **on the list**.
> 
> Sorry, what is the list? And I was just wondering since it hasn't been reviewed for a few months, and it's quite important for stuff like `run0`.

The list of reviews to do. Though I am aware of the advantage of this convenience-level PR, still, security-level tasks have higher priority. Be sure that the polkit team is aware of all the pull requests, pinging won't make them reviewed faster (this is a general, definitely not a personal remark). 

If this commit is important to you, there's always a COPR build available or pull it into your own build. It will not be available otherwise until the next release (somewhere in summer) anyway.

jedenastka:
Usability level questions:

> Unlike sudo, we also do not refresh the countdown on each use.

1. Why not?
2. Could that be made configurable as part of this MR, or should that be a new one?

I personally often rely on this for doing longer chains of actions with sudo, interleaved with regular commands as my user. I probably could work with this anyways for the most part, but it is convenient.

bluca:
There's no caching at all right now. One thing at a time.

jedenastka:
Got it.

jrybar-rh:
Note to self: Again, this change requires fix in SELinux policies to work on Fedora and other OSes with SELinux enabled.

oech3:
What is the most simple conf to use this? I checked this is part of `pokit 126`.

jrybar-rh:
> What is the most simple conf to use this? I checked this is part of `pokit 126`.

I don't think it is. No conf needed.

jrybar-rh:
> Empty `/etc/polkit-1/rules.d` does not skip 2nd auth. What should `YES` of
> 
> ```
> polkit.addRule(function(action, subject) {
>     if (subject.isInGroup("wheel")) {
>         return polkit.Result.YES;
>     }
> });
> ```
> 
> replaced with?

I'm almost positive this merge request is not part of released version of polkit yet.

oech3:
I see. Thankyou.

PR #535, opened 2024-12-17
Explicitly use chdir("/") instead of relying on ~user being set properly

bluca:
On non-systemd systems if the user is changed manually, polkitd will also change directory. The homedir of the user might not be set correctly, so just change to /.

This is a no-op on systemd systems, as the user is set in the unit file, so this code never runs.

### DIFF
--- a/src/polkitbackend/polkitd.c
+++ b/src/polkitbackend/polkitd.c
@@ -184,10 +184,10 @@ become_user (const gchar  *user,
       goto out;
     }
 
-  if (chdir (pw->pw_dir) != 0)
+  if (chdir ("/") != 0)
     {
       g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
-                   "Error changing to home directory %s: %m",
+                   "Error changing to root directory %s: %m",
                    pw->pw_dir);
       goto out;
     }


mbiebl:
It might not be set correctly or explicity use something like `/nonexistent` as in https://salsa.debian.org/utopia-team/polkit/-/blob/master/debian/polkitd.postinst#L61

bluca:
> It might not be set correctly or explicity use something like `/nonexistent` as in https://salsa.debian.org/utopia-team/polkit/-/blob/master/debian/polkitd.postinst#L61

Yep

@jrybar-rh any chance we can get this merged too, please? Thanks!

jrybar-rh:
> > It might not be set correctly or explicity use something like `/nonexistent` as in https://salsa.debian.org/utopia-team/polkit/-/blob/master/debian/polkitd.postinst#L61
> 
> Yep
> 
> @jrybar-rh any chance we can get this merged too, please? Thanks!

Yes, I wanted to give you chance to react. This is another strictly Debian thing. So it's up to you guys whether or not it's a desired change.
FYI, @vmihalko, @mrc0mmand and I are on Xmas vacation since Monday, so next work on polkit project is planned after New Year.
Which leads me to: Merry Christmas, happy and joyful relax and I'm looking forward to further cooperation next year!

bluca:
> > > It might not be set correctly or explicity use something like `/nonexistent` as in https://salsa.debian.org/utopia-team/polkit/-/blob/master/debian/polkitd.postinst#L61
> > 
> > 
> > Yep
> > @jrybar-rh any chance we can get this merged too, please? Thanks!
> 
> Yes, I wanted to give you chance to react. This is another strictly Debian thing. So it's up to you guys whether or not it's a desired change.

Yeah we carried this patch for years, it's a fallback for the non-systemd case.

> FYI, @vmihalko, @mrc0mmand and I are on Xmas vacation since Monday, so next work on polkit project is planned after New Year. Which leads me to: Merry Christmas, happy and joyful relax and I'm looking forward to further cooperation next year!

Merry christmas!

Issue #539, opened 2025-01-14
polkit-126 fails to build with `-Dsession_tracking=elogind`


Polynomial-C:
**Describe the bug**
`polkit-126` fails to build when using elogind as session tracker:
```
x86_64-pc-linux-gnu-gcc -Isrc/polkitbackend/polkitd.p -Isrc/polkitbackend -I../polkit-126/src/polkitbackend -I. -I../polkit-126 -Isrc -I../polkit-126/src -Isrc/polkit -I/usr/lib64/libffi/include -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include -I/usr/include/libmount -I/usr/include/uuid -I/usr/include/json-c -I/usr/include/blkid -I/usr/include/gio-unix-2.0 -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=c99 -D_GNU_SOURCE -include config.h -Wno-format-y2k -Wno-declaration-after-statement -Wno-unused-result -march=native -mtune=native -O2 -pipe -pthread '-DG_LOG_DOMAIN="polkitd-1"' -DPOLKIT_BACKEND_I_KNOW_API_IS_SUBJECT_TO_CHANGE -MD -MQ src/polkitbackend/polkitd.p/polkitd.c.o -MF src/polkitbackend/polkitd.p/polkitd.c.o.d -o src/polkitbackend/polkitd.p/polkitd.c.o -c ../polkit-126/src/polkitbackend/polkitd.c
../polkit-126/src/polkitbackend/polkitd.c:34:12: fatal error: systemd/sd-daemon.h: No such file or directory
   34 | #  include <systemd/sd-daemon.h>
      |            ^~~~~~~~~~~~~~~~~~~~~
compilation terminated.
ninja: build stopped: subcommand failed.
```

As you can see, the compiler command lacks the important `-I/usr/include/elogind` and that is where `systemd/sd-daemon.h` can be found when `elogind` is being used.

**To Reproduce**
1. Have a system with `elogind` as logind provider (no systemd!)
2. Compile the project with `-Dsession_tracking=elogind`
3. Wait for the above error message

**Expected behavior**
Build succeeds like with previous releases ;)

**Desktop (please complete the following information):**
 - OS (including version): Gentoo Linux
 - Version of polkit: 126

**Additional context**
I am not really experienced with meson build system but right now I use this workaround to make polkit-126 build succeed:
```
[...]

PR #548, opened 2025-02-04
ci: build & run unit tests on Alpine as well

mrc0mmand:
Alpine uses musl + elogind, which should provide some additional
coverage, especially for issues like https://github.com/polkit-org/polkit/issues/539.

---

\+ a couple of fixes for build warnings/errors when building with musl.

### DIFF
--- a/.github/workflows/ci.sh
+++ b/.github/workflows/ci.sh
@@ -6,12 +6,13 @@ set -o pipefail
 #   - drop -Wno-deprecated-declarations
 
 PHASE="${1:?}"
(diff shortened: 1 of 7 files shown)


mrc0mmand:
(All F42 fails are expected, there's ongoing branching in Fedora and it'll break stuff until it's complete, see https://pagure.io/fedora-infrastructure/issue/12384)

jrybar-rh (approved):

vmihalko (approved):

PR #551, opened 2025-02-12
ci: install newer meson for the Coverity job as well

mrc0mmand:
Since it also runs on Ubuntu 24.04 which doesn't have new-enough meson.

Follow-up for 810b5dd863dbbb0a19c3613a1f5dd55c2e8ac886.

---

(Note: this is intentionally _not_ opened from a fork branch, so I can test the Coverity job indeed works as it needs access to repo tokens)


mrc0mmand:
Coverity seems to be happy once again: https://github.com/polkit-org/polkit/actions/runs/13281808156/job/37081630417?pr=551

PR #553, opened 2025-02-19
Fix musl compilation error on Alpine

jrybar-rh:
Feature-test macro _XOPEN_SOURCE is said to be an equivalent to _POSIX_C_SOURCE and apparently is defined on Alpine.


jrybar-rh:
or it isn't. :roll_eyes:

PR #557, opened 2025-04-02
polkit-agent-helper: set CollectMode=inactive-or-failed

bluca:
Ensure that failed helper runs (eg: auth failure) do not leave units behind to accumulate

Follow-up for c007940054392b67b84b6044dda8a215040d8eb5

### DIFF
--- a/data/polkit-agent-helper@.service.in
+++ b/data/polkit-agent-helper@.service.in
@@ -1,6 +1,7 @@
 [Unit]
 Description=Authorization Manager Agent Helper
 Documentation=man:polkit(8)
+CollectMode=inactive-or-failed
 
 [Service]
 Type=oneshot


jrybar-rh:
There seem to be two issues with this approach:
1. it still leaves the system in degraded state for who knows how long (after local discussion with systemd guys, GC is now event-based instead of time-based). Failed password is not fundamentally an error of the program. The helper finishes correctly in reaction to wrong user interaction. This should not cause unit failure and definitely not send the system to degraded state. Imagine the red flags this raises in some setups. "Degraded? What's happening? Wrong password? Just this?"
2. GC'ing failed service may clear the system of wrong passwords, but also masks potential real problems with the service. There are many more events with return code 1 in the helper program. Say we cannot init a pam stack. This gets GC'ed few seconds later. Still in journal, but what's the motivation to see journal when system looks ok.

@lnykryn suggested a directive I was looking for, so modifying return codes seemed necessary yet logical: #558

Issue #561, opened 2025-05-13
[RFE] Consider setting $SUDO_UID in pkexec, just like sudo, run0, sudo-rs all do


poettering:
I'd like to suggest that pkexec sets the $SUDO_UID env var to the originating, numeric UID during its security transition, the same way as classic "sudo" do it, but also how systemd's run0 does it, and sudo-rs.

I think at this point we should just consider $SUDO_UID a generic API, that even while it was introduced by sudo is generic enough to be universal. 

There are a multitude of tools that look at the thing. Insider the systemd project we want to check for that too, to disable certain functionality when some of our tools are invoked in a sudo or sudo-equivalent context. This already works fine with sudo, sudo-rs, and run0. But currently does not work with pkexec. I think it would be great if it did.

Yes, our systemd code could check for $PKEXEC_UID in addition to $SUDO_UID, but we really don't want to list every possible implementation like this. I think it would be much nicer if sudo-like tools would just provide the modicum of compat on this and all set $SUDO_UID.

(of course $SUDO_GID, and maybe $SUDO_USER should be initialized too, but to me those don't matter much)

Thank you for considering.

PR #562, opened 2025-05-13
pkexec: also se $SUDO_UID/$SUDO_GID for compat with sudo

poettering:
Various tools (including some in systemd) want to slightly change behaviour in case they detect they are invoked in a sudo or sudo-like context (for example disable certain security-sensitive interfaces). They currently mostly check $USER_UID to detect this situation, which will work for at least the sudo, run0, sudo-rs implementation of a privilege elevation tool. It doesn't work for pkexec however.

Let's change this, and for compatibility with sudo simply set the these two env vars too. This would increase compatibility for everyone in a simple, generic way.

Yes, it would be great if the env var would be more generically named than $SUDO_something, but I think at this point it shouldn't really matter, it's just a string.

Fixes: #561

### DIFF
--- a/src/programs/pkexec.c
+++ b/src/programs/pkexec.c
@@ -926,6 +926,12 @@ main (int argc, char *argv[])
 
   s = g_strdup_printf ("%d", getuid ());
   g_ptr_array_add (saved_env, g_strdup ("PKEXEC_UID"));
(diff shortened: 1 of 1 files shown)


keszybz (approved):
LGTM.

PR #563, opened 2025-05-14
Fix rapid leaking of large amounts of file descriptors

msizanoen1:
## Summary
Prevents polkit from leaking massive amounts (~10-20) of pidfds every invocation.

See individual commit messages for more details.

Tested using `run0` and a bpftrace script to monitor pidfd open/close for one invocation.

bpftrace script:
```bpftrace
tracepoint:syscalls:sys_exit_pidfd_open /comm == "polkitd" && args->ret > 0/ {
        @opener[args->ret] = ustack;
        @ispfd[args->ret] = 1;
        @ocnt[ustack]++;
}

tracepoint:syscalls:sys_enter_close /comm == "polkitd" && @ispfd[args->fd] == 1/ {
        if (--@ocnt[@opener[args->fd]] == 0) {
                delete(@ocnt[@opener[args->fd]]);
        }
        delete(@opener[args->fd]);
        delete(@ispfd[args->fd]);
}

END { clear(@ispfd); }
```

Before patch:
```
Attaching 3 probes...



@ocnt[
        syscall+29
        polkit_unix_process_set_pidfd+141
        object_set_property+169
        g_object_new_internal.part.0+360
        g_object_new_valist+1595
        g_object_new+159
        polkit_unix_process_new_pidfd+70
        polkit_system_bus_name_get_process_sync+224
        _polkit_subject_get_cmdline+147
        check_authorization_challenge_cb+148
        authentication_agent_begin_cb+164
        g_task_return_now+44
        g_task_return+304
        reply_cb+151
        g_task_return_now+44
        g_task_return+304
        g_dbus_connection_call_done+354
        g_task_return_now+44
        complete_in_idle_cb+21
        g_idle_dispatch+45
        g_main_context_dispatch_unlocked.lto_priv.0+464
        g_main_context_iterate_unlocked.isra.0+728
        g_main_loop_run+127
        main+1097
        __libc_start_call_main+117
        __libc_start_main_alias_1+136
        _start+37
]: 1
@ocnt[
        syscall+29
        polkit_unix_process_set_pidfd+141
        object_set_property+169
        g_object_new_internal.part.0+360
        g_object_new_valist+1595
        g_object_new+159
        polkit_unix_process_new_pidfd+70
        polkit_system_bus_name_get_process_sync+224
        polkit_backend_interactive_authority_check_authorization+2145
        server_handle_method_call+569
        call_in_idle_cb+278
        g_idle_dispatch+45
        g_main_context_dispatch_unlocked.lto_priv.0+464
        g_main_context_iterate_unlocked.isra.0+728
        g_main_loop_run+127
        main+1097
        __libc_start_call_main+117
        __libc_start_main_alias_1+136
        _start+37
]: 1
[lots of similar entries]
@ocnt[
        syscall+29
        polkit_unix_process_set_pidfd+141
        object_set_property+169
        g_object_new_internal.part.0+360
        g_object_new_valist+1595
        g_object_new+159
        polkit_unix_process_new_pidfd+70
        polkit_system_bus_name_get_process_sync+224
        convert_temporary_authorization_subject+85
        temporary_authorization_store_has_authorization+97
        check_authorization_sync+433
        check_authorization_sync+1453
        polkit_backend_interactive_authority_check_authorization+881
        server_handle_method_call+569
        call_in_idle_cb+278
        g_idle_dispatch+45
        g_main_context_dispatch_unlocked.lto_priv.0+464
        g_main_context_iterate_unlocked.isra.0+728
        g_main_loop_run+127
        main+1097
        __libc_start_call_main+117
        __libc_start_main_alias_1+136
        _start+37
]: 1
@ocnt[
        syscall+29
        polkit_unix_process_set_pidfd+141
        object_set_property+169
        g_object_new_internal.part.0+360
        g_object_new_valist+1595
        g_object_new+159
        polkit_unix_process_new_pidfd+70
        polkit_system_bus_name_get_process_sync+224
        push_subject+934
        polkit_backend_common_js_authority_check_authorization_sync+292
        check_authorization_sync+385
        polkit_backend_interactive_authority_check_authorization+881
        server_handle_method_call+569
        call_in_idle_cb+278
        g_idle_dispatch+45
        g_main_context_dispatch_unlocked.lto_priv.0+464
        g_main_context_iterate_unlocked.isra.0+728
        g_main_loop_run+127
        main+1097
        __libc_start_call_main+117
        __libc_start_main_alias_1+136
        _start+37
]: 25
@ocnt[
        syscall+29
        polkit_unix_process_set_pidfd+141
        object_set_property+169
        g_object_new_internal.part.0+360
        g_object_new_valist+1595
        g_object_new+159
        polkit_unix_process_new_pidfd+70
        polkit_system_bus_name_get_process_sync+224
        polkit_backend_session_monitor_get_session_for_subject.constprop.0+341
        check_authorization_sync+197
        polkit_backend_interactive_authority_check_authorization+881
        server_handle_method_call+569
        call_in_idle_cb+278
        g_idle_dispatch+45
        g_main_context_dispatch_unlocked.lto_priv.0+464
        g_main_context_iterate_unlocked.isra.0+728
        g_main_loop_run+127
        main+1097
        __libc_start_call_main+117
        __libc_start_main_alias_1+136
        _start+37
]: 25
@opener[42]: 
        syscall+29
        polkit_unix_process_set_pidfd+141
        object_set_property+169
        g_object_new_internal.part.0+360
        g_object_new_valist+1595
        g_object_new+159
        polkit_unix_process_new_pidfd+70
        polkit_system_bus_name_get_process_sync+224
        polkit_backend_session_monitor_get_session_for_subject.constprop.0+341
        check_authorization_sync+197
        polkit_backend_interactive_authority_check_authorization+881
        server_handle_method_call+569
        call_in_idle_cb+278
        g_idle_dispatch+45
        g_main_context_dispatch_unlocked.lto_priv.0+464
        g_main_context_iterate_unlocked.isra.0+728
        g_main_loop_run+127
        main+1097
        __libc_start_call_main+117
        __libc_start_main_alias_1+136
        _start+37
[lots of similar entries]
@opener[106]: 
        syscall+29
        polkit_unix_process_set_pidfd+141
        object_set_property+169
        g_object_new_internal.part.0+360
        g_object_new_valist+1595
        g_object_new+159
        polkit_unix_process_new_pidfd+70
        polkit_system_bus_name_get_process_sync+224
        push_subject+934
        polkit_backend_common_js_authority_check_authorization_sync+292
        check_authorization_sync+385
        polkit_backend_interactive_authority_check_authorization+881
        server_handle_method_call+569
        call_in_idle_cb+278
        g_idle_dispatch+45
        g_main_context_dispatch_unlocked.lto_priv.0+464
        g_main_context_iterate_unlocked.isra.0+728
        g_main_loop_run+127
        main+1097
        __libc_start_call_main+117
        __libc_start_main_alias_1+136
        _start+37

```

After patch:
```
Attaching 3 probes...



@ocnt[
        syscall+29
        polkit_unix_process_set_pidfd+148
        object_set_property+169
        g_object_new_internal.part.0+360
        g_object_new_valist+1595
        g_object_new+159
        polkit_unix_process_new_pidfd+70
        polkit_system_bus_name_get_process_sync+224
        convert_temporary_authorization_subject+85
        check_authorization_challenge_cb+648
        authentication_agent_begin_cb+164
        g_task_return_now+44
        g_task_return+304
        reply_cb+151
        g_task_return_now+44
        g_task_return+304
        g_dbus_connection_call_done+354
        g_task_return_now+44
        complete_in_idle_cb+21
        g_idle_dispatch+45
        g_main_context_dispatch_unlocked.lto_priv.0+464
        g_main_context_iterate_unlocked.isra.0+728
        g_main_loop_run+127
        main+1097
        __libc_start_call_main+117
        __libc_start_main_alias_1+136
        _start+37
]: 1
@opener[10]: 
        syscall+29
        polkit_unix_process_set_pidfd+148
        object_set_property+169
        g_object_new_internal.part.0+360
        g_object_new_valist+1595
        g_object_new+159
        polkit_unix_process_new_pidfd+70
        polkit_system_bus_name_get_process_sync+224
        convert_temporary_authorization_subject+85
        check_authorization_challenge_cb+648
        authentication_agent_begin_cb+164
        g_task_return_now+44
        g_task_return+304
        reply_cb+151
        g_task_return_now+44
        g_task_return+304
        g_dbus_connection_call_done+354
        g_task_return_now+44
        complete_in_idle_cb+21
        g_idle_dispatch+45
        g_main_context_dispatch_unlocked.lto_priv.0+464
        g_main_context_iterate_unlocked.isra.0+728
        g_main_loop_run+127
        main+1097
        __libc_start_call_main+117
        __libc_start_main_alias_1+136
        _start+37

```

From https://docs.gtk.org/gobject/vfunc.Object.constructed.html:
> The constructed function is called by g_object_new() as the final step of the object creation process. **At the point of the call, all construction properties have been set on the object.** The purpose of this call is to allow for object initialisation steps that can only be performed after construction properties have been set. constructed implementors should chain up to the constructed call of their parent class to allow it to complete its initialisation.

### DIFF
--- a/src/polkit/polkitunixprocess.c
+++ b/src/polkit/polkitunixprocess.c
@@ -1065,7 +1065,7 @@ polkit_unix_process_hash (PolkitSubject *subject)
 {
   PolkitUnixProcess *process = POLKIT_UNIX_PROCESS (subject);
 
-  return g_direct_hash (GSIZE_TO_POINTER ((polkit_unix_process_get_pid(process) + process->start_time))) ;
+  return g_direct_hash (GSIZE_TO_POINTER ((process->pid + process->start_time))) ;
 }
 
 static gboolean
@@ -1076,6 +1076,9 @@ polkit_unix_process_equal (PolkitSubject *a,
   PolkitUnixProcess *process_b;
   gint pid_a, pid_b;
   gint pidfd_a, pidfd_b;
+
+  if (a == b)
+    return TRUE;
 
   process_a = POLKIT_UNIX_PROCESS (a);
   process_b = POLKIT_UNIX_PROCESS (b);
--- a/src/polkit/polkitunixprocess.c
+++ b/src/polkit/polkitunixprocess.c
@@ -399,12 +399,15 @@ polkit_unix_process_constructed (GObject *object)
         GError *error;
         error = NULL;
 
-        process->ppidfd = get_ppidfd_for_pidfd (process->pidfd, &error);
-        if (error != NULL)
+        if (process->ppidfd < 0)
           {
-            g_error_free (error);
-            error = NULL;
-            process->ppidfd = -1;
+            process->ppidfd = get_ppidfd_for_pidfd (process->pidfd, &error);
+            if (error != NULL)
+              {
+                g_error_free (error);
+                error = NULL;
+                process->ppidfd = -1;
+              }
           }
 
         process->cgroupid = get_cgroupid_for_pidfd (process->pidfd, &error);
--- a/src/polkit/polkitunixprocess.c
+++ b/src/polkit/polkitunixprocess.c
@@ -840,7 +840,7 @@ polkit_unix_process_set_pid (PolkitUnixProcess *process,
     }
   if (pid > 0)
     {
-      gint pidfd = (int) syscall (SYS_pidfd_open, process->pid, 0);
+      gint pidfd = (int) syscall (SYS_pidfd_open, pid, 0);
       if (pidfd >= 0)
         {
           GError *error;
--- a/src/polkit/polkitunixprocess.c
+++ b/src/polkit/polkitunixprocess.c
@@ -165,6 +165,7 @@ struct _PolkitUnixProcess
   guint ctty;
   gboolean pidfd_is_safe;
   GArray *gids;
+  gboolean constructed;
 };
 
 struct _PolkitUnixProcessClass
@@ -437,6 +438,8 @@ polkit_unix_process_constructed (GObject *object)
           g_error_free (error);
         }
     }
+
+  process->constructed = TRUE;
 
   if (G_OBJECT_CLASS (polkit_unix_process_parent_class)->constructed != NULL)
     G_OBJECT_CLASS (polkit_unix_process_parent_class)->constructed (object);
@@ -832,35 +835,38 @@ polkit_unix_process_set_pid (PolkitUnixProcess *process,
   g_return_if_fail (POLKIT_IS_UNIX_PROCESS (process));
 
 #ifdef HAVE_PIDFD_OPEN
-  if (process->pidfd >= 0)
+  if (process->constructed)
     {
-      close (process->pidfd);
-      process->pidfd = -1;
-      process->pidfd_is_safe = FALSE;
-    }
-  if (pid > 0)
-    {
-      gint pidfd = (int) syscall (SYS_pidfd_open, pid, 0);
-      if (pidfd >= 0)
+      if (process->pidfd >= 0)
         {
-          GError *error;
-          error = NULL;
-
+          close (process->pidfd);
+          process->pidfd = -1;
           process->pidfd_is_safe = FALSE;
-          process->pidfd = pidfd;
-          process->pid = 0;
-          if (process->ppidfd >= 0)
+        }
+      if (pid > 0)
+        {
+          gint pidfd = (int) syscall (SYS_pidfd_open, pid, 0);
+          if (pidfd >= 0)
             {
-              close (process->ppidfd);
-              process->ppidfd = -1;
+              GError *error;
+              error = NULL;
+
+              process->pidfd_is_safe = FALSE;
+              process->pidfd = pidfd;
+              process->pid = 0;
+              if (process->ppidfd >= 0)
+                {
+                  close (process->ppidfd);
+                  process->ppidfd = -1;
+                }
+              process->ppidfd = get_ppidfd_for_pidfd (process->pidfd, &error);
+              if (error != NULL)
+                {
+                  g_error_free (error);
+                  process->ppidfd = -1;
+                }
+              return;
             }
-          process->ppidfd = get_ppidfd_for_pidfd (process->pidfd, &error);
-          if (error != NULL)
-            {
-              g_error_free (error);
-              process->ppidfd = -1;
-            }
-          return;
         }
     }
 #endif /* HAVE_PIDFD_OPEN */


src/polkit/polkitunixprocess.c
```diff
@@ -376,37 +376,19 @@ polkit_unix_process_constructed (GObject *object)
...
-  /* sets pidfd, start_time and uid in case they are unset */
+  /* sets start_time and uid in case they are unset */
 
   /* We didn't open it ourselves here, so we must have got it
    * from D-Bus, mark it as safe to use */
   if (process->pidfd >= 0)
     process->pidfd_is_safe = TRUE;
 
 #ifdef HAVE_PIDFD_OPEN
-  if (process->pid > 0 && process->pidfd < 0)
```
bluca, line 387: This doesn't look right, it depends on how the object is created whether it's set or not. There's a check to see if it was already set, so just leave this as-is

src/polkit/polkitunixprocess.c
```diff
@@ -376,37 +376,19 @@ polkit_unix_process_constructed (GObject *object)
...
-          process->pid = 0;
-        }
-    }
-
     if (process->pidfd >= 0)
       {
         GError *error;
         error = NULL;
 
-        process->ppidfd = get_ppidfd_for_pidfd (process->pidfd, &error);
```
bluca, line 402: Same as above, this cannot be dropped. Instead add the same check as above to see if it was already set

src/polkit/polkitunixprocess.c
```diff
@@ -932,6 +927,7 @@ polkit_unix_process_set_pidfd (PolkitUnixProcess *process,
       GError *error;
       error = NULL;
 
+      process->pidfd_is_safe = TRUE;
```
bluca, line 930: This doesn't look right either, the purpose of this is very specifically to know if it came from the dbus peer or not, and it's lost with this change. Please drop it.

msizanoen1: The check in the `->constructed()` callback never worked in the first place because the setter for the "pid" property also sets the pidfd field, which means if the process was constructed using a "pid" property then by the time `->constructed()` is invoked the pidfd wouls already be set and the pidfd is then incorrectly marked as safe.

Please see https://docs.gtk.org/gobject/vfunc.Object.constructed.html as well as the commit messages for the changes.

bluca: What tests have you performed to verify those claims, in details?

msizanoen1: See the code for `polkit_unix_process_set_pid` which is called whenever the "pid" GObject property is set: https://github.com/polkit-org/polkit/blob/894e11120d096b469677830aa94d40e309c22fe6/src/polkit/polkitunixprocess.c#L818-L866

According to the GObject docs linked above, it should be invoked before `->construct()` if the object was constructed with a "pid" property.

msizanoen1: There's another problem however: all `G_PARAM_CONSTRUCT` properties are set with their default values during construction, and there's no guarantees about the orders in which properties are set. This means if the object was constructed using a pidfd, both the "pid" and "pidfd" properties are set in unspecified order to 0 and the actual pidfd respectively. The same is true for constructing the object from a PID, where the "pidfd" property is also set to -1 which may cause the pidfd set by the "pid" setter to be overwritten. I will push changes to address this soon.

bluca: Sorry, but the question was intentionally about which tests were performed to verify those claims. I assume this means no tests were done?

I asked about tests because a very simple check shows the claim to be incorrect. The flag is set correctly when the dbus daemon sends a pidfd, and it's not set when it doesn't and it's opened locally instead, as expected and documented.

So once again, please drop this change, and part of the other too, and just do add a check for the ppidfd existence, which will solve any issue with potentially duplicated FDs. Thanks.

msizanoen1: https://github.com/polkit-org/polkit/blob/894e11120d096b469677830aa94d40e309c22fe6/src/polkit/polkitunixprocess.c#L840

Seems like `polkit_unix_process_set_pid` is using `process->pid` (instead of the `pid` parameter) for `pidfd_open` even if it hasn't been initialized yet. This would obviously result in no pidfd being opened (since 0 is not a valid PID):
```
pidfd_open(0, 0)                        = -1 EINVAL (Invalid argument)
```

Would it be acceptable to just remove pidfd handling from `set_pid` entirely since this code path apparently never worked in the first place?

bluca: That is needed in some cases, depending on how it's called. Please just fix that issue by changing from process->pid to pid.

src/polkit/polkitunixprocess.c
```diff
@@ -1065,7 +1065,7 @@ polkit_unix_process_hash (PolkitSubject *subject)
 {
   PolkitUnixProcess *process = POLKIT_UNIX_PROCESS (subject);
 
-  return g_direct_hash (GSIZE_TO_POINTER ((polkit_unix_process_get_pid(process) + process->start_time))) ;
+  return g_direct_hash (GSIZE_TO_POINTER ((process->pid + process->start_time))) ;
```
bluca, line 1068: This is very wrong, the pid _must_ be resolved, as the entire point is to check if it did change, and if it did, it's no longer the same

src/polkit/polkitunixprocess.c
```diff
@@ -1077,6 +1077,9 @@ polkit_unix_process_equal (PolkitSubject *a,
   gint pid_a, pid_b;
   gint pidfd_a, pidfd_b;
 
+  if (a == b)
```
bluca, line 1080: As above, this is not right, the pids must be resolved

src/polkit/polkitunixprocess.c
```diff
@@ -165,6 +165,7 @@ struct _PolkitUnixProcess
   guint ctty;
   gboolean pidfd_is_safe;
   GArray *gids;
+  gboolean constructed;
```
bluca, line 168: This is pointless and confusing. The object is either created with a pid or a pidfd, not both. Please drop this commit, and the next one. The first two commits are good.

msizanoen1: All `G_PARAM_CONSTRUCT` properties means all of them, including default values. If you don't specify a "pid" during construction GLib will act as if "pid" is set to 0 (vice versa with "pidfd" except it's -1). This case must be handled.

src/polkit/polkitunixprocess.c
msizanoen1: The hash doesn't need to be unique - simply unique enough to not cause performance degradation. The hash value must be constant throughout the object lifecycle for https://github.com/polkit-org/polkit/blob/894e11120d096b469677830aa94d40e309c22fe6/src/polkitbackend/polkitbackendinteractiveauthority.c#L2841 to work.

src/polkit/polkitunixprocess.c
msizanoen1: I can't see the reason why always treating an object as equal to itself is an issue? This is needed so you can remove the object if it's being used as a key to a hash table e.g. https://github.com/polkit-org/polkit/blob/894e11120d096b469677830aa94d40e309c22fe6/src/polkitbackend/polkitbackendinteractiveauthority.c#L2841

bluca: No, this reintroduces a security issue. The pid must always be resolved. Please drop this commit.

src/polkit/polkitunixprocess.c
bluca: No, this introduces a security risk. The pid must be resolved. Please drop this commit.

src/polkit/polkitunixprocess.c
bluca: It is already handled. This serves no purpose, please drop this commit.

src/polkit/polkitunixprocess.c
msizanoen1: Assuming the security issue in question is the pid reuse problem that was resolved with the introduction of pidfd support, wouldn't any `PolkitUnixProcess` objects created later be a different object and thus have a different address and would not pass this equality check? The `pid` itself is never compared, but the object itself is.

src/polkit/polkitunixprocess.c
msizanoen1: The hash being used here is for a hash table. There should be no way for this to create a security issue.

src/polkit/polkitunixprocess.c
msizanoen1: Only for the case where `pid` is set before `pidfd` during construction. Since the order in which properties are set during construction is entirely arbitrary and unspecified, there is no guarantee that this ordering behavior will be the same in future GLib versions, which may set `pid` after `pidfd` and break construction of `PolkitUnixProcess`es from pidfds.

src/polkit/polkitunixprocess.c
msizanoen1: Note that it's assumed here that the `pid` field will be 0 if `pidfd` is set, which means `start_time` alone will be used for hashing.

src/polkit/polkitunixprocess.c
bluca: No. The pid must always be resolved, all the time, for any purpose, in any situation. Please drop this commit.

If you are trying to solve some specific use case, please provide exact details of what real-world scenario is being affected, how it was tested, and results before/after.
Otherwise please just drop the 3rd and 4th commits. Thanks.

msizanoen1: Reproduced on Fedora 42, using `bpftrace` and the `run0` utility from systemd.

Can also be verified by repeatedly running `run0 ls -l /proc/$(pidof polkitd)/fd`. It can be directly observed that a new pidfd is leaked for every invocation (with 2 first commits applied only)

`bpftrace` script to demonstrate change in hash value causing failure to remove the object from hash table:
```bpftrace
uprobe:libpolkit-gobject-1:polkit_unix_process_equal /arg0 == arg1/ {
	@pending_compare_self[tid] = arg0;
}

/* note: this is unreachable since due to how hash tables work the hashes need to be equal for this to be reachable in the first place */
/* not relevant for now */
uretprobe:libpolkit-gobject-1:polkit_unix_process_equal /@pending_compare_self[tid] != 0/ {
	if (retval == 0) {
		printf("%d(%s)object(%p) not equal to itself, stacktrace=%s\n", pid, comm, @pending_compare_self[pid], ustack);
	}
	delete(@pending_compare_self[tid]);
}

uprobe:libpolkit-gobject-1:polkit_unix_process_init {
	@interesting[pid, arg0] = 1;
}

uprobe:libpolkit-gobject-1:polkit_unix_process_finalize {
	delete(@interesting[pid, arg0]);
}

uprobe:libpolkit-gobject-1:polkit_unix_process_hash {
	@pending_hash[tid] = arg0;
}

uretprobe:libpolkit-gobject-1:polkit_unix_process_hash /@pending_hash[tid] != 0/ {
	$arg0 = @pending_hash[tid];
	$prev_hash = @hashof[pid, $arg0];
	if ($prev_hash != 0 && $prev_hash != retval) {
		printf("%d(%s)object(%p) hash changed prev=%d now=%d stacktrace=%s\n", pid, comm, $arg0, $prev_hash, retval, ustack);
	}
	@hashof[pid, $arg0] = retval;
	delete(@pending_hash[tid]);
}

uprobe:libpolkit-gobject-1:polkit_unix_process_finalize /@hashof[pid, arg0] != 0/ {
	delete(@hashof[pid, arg0]);
}

uprobe:libglib-2.0:g_hash_table_remove {
	@pending_rm[tid] = arg1;
}

uretprobe:libglib-2.0:g_hash_table_remove /@pending_rm[tid] != 0/ {
	$arg1 = @pending_rm[tid];
	if (retval == 0 && @interesting[pid, $arg1]) {
		printf("%d(%s)object(%p) failed to be removed from a hash table, stacktrace=%s\n", pid, comm, $arg1, ustack);
	}
	delete(@pending_rm[tid]);
}
```

`bpftrace` output for a `run0 ls -l /proc/$(pidof polkitd)/fd` invocation:
```
Attaching 9 probes...
38257(polkitd)object(0x56078e8d4cb0) hash changed prev=8768385 now=8715121 stacktrace=
        g_hash_table_remove+45
        on_name_owner_changed_signal+1236
        emit_signal_instance_in_idle_cb+128
        g_idle_dispatch+45
        g_main_context_dispatch_unlocked.lto_priv.0+464
        g_main_context_iterate_unlocked.isra.0+728
        g_main_loop_run+127
        main+1023
        __libc_start_call_main+117
        __libc_start_main_alias_1+136
        _start+37

38257(polkitd)object(0x56078e8d4cb0) failed to be removed from a hash table, stacktrace=
        on_name_owner_changed_signal+1236
        emit_signal_instance_in_idle_cb+128
        g_idle_dispatch+45
        g_main_context_dispatch_unlocked.lto_priv.0+464
        g_main_context_iterate_unlocked.isra.0+728
        g_main_loop_run+127
        main+1023
        __libc_start_call_main+117
        __libc_start_main_alias_1+136
        _start+37



@hashof[38257, 94590456384688]: 8715121
@interesting[53265, 94200297301600]: 1
@interesting[38257, 94590456384688]: 1
@interesting[38257, 94590456992704]: 1




```

bluca: > Can also be verified by repeatedly running run0 ls -l /proc/$(pidof polkitd)/fd. It can be directly observed that a new pidfd is leaked for every invocation (the first commit applied only - commit 2 is broken without commit 3, see below)

No, that is not the case, the issue is fixed by the first commit, which is good as already mentioned. After that, only 2 pidfds are held per cached authentication, as expected, this can be seen trivially from fdinfo. This commit is not necessary and in fact it's damaging as it reintroduces pid reuse vulnerabilities, so please drop it. Thanks.

src/polkit/polkitunixprocess.c
bluca: Once again, the actual reported issue is fixed by the first commit, the rest is working as intended. Please drop this commit, the first two are good and can be merged. Thanks.

bluca:
Why did you close this? As mentioned multiple times, the first two commits were good and correct bug fixes, and could have been merged without issues, as they solved a bug introduced by 7d44d62a02f090a5a71c4ca10d58c7c2b54881eb

msizanoen1:
If you try applying only the first 2 commits without the third, you will see that `polkit_unix_process_new` as well as the variant that also takes in a UID no longer works and always return 0 when you try to query for its PID. Commit 2 does not, in fact, work without commit 3.

I will send a new PR with only the first commit that's not known to break anything, and unless you can ask another maintainer who is more well-versed in GLib/GObject (since the later 2 commits are rather heavy in obscure GLib details) to review the other ones that will be the only part of this PR that will be merged.

bluca:
_if_ the process is gone then yes, it doesn't match anymore. That's not only good, it's necessary. Two polkit subjects must not evaluate as equal under any circumstances just because at some point during their lifecycles they had the same pid. The pid must always be resolved at the point of use, for any purpose, otherwise the subject is vulnerable to reuse attacks. This is regardless of any glib details, as it's a polkit-specific detail, and it's of paramount importance, otherwise we'll be right back at making it impossible to use polkit for anything other than D-Bus peers. We already rely on it heavily for varlink IPC too, so regressions on that front cannot be risked, sorry.

If you want to only send a single fix that's of course entirely ok and up to you. You can also force push to this same PR if you wish, rather than creating a different one, that's also up to you. Thanks for the fix(es).

PR #565, opened 2025-05-17
PolkitUnixProcess: Don't overwrite ppidfd in constructed callback

msizanoen1:
The parent pidfd might already be set by the "pidfd" property setter during object construction. If that's the case, we shouldn't try to open the parent pidfd again.


bluca (approved):
Thanks for spotting this and providing a fix, tested locally and looks good to me

Bug was introduced by 7d44d62a02f090a5a71c4ca10d58c7c2b54881eb

PR #579, opened 2025-08-07
Give uid to rules

Rosentti:
## Summary
Added user id to the `subject` for use by rules.
Usecase: Allow rules to determine if a subject is a "real user" (user with >= 1000 uid)


jrybar-rh:
This shall be merged in #616 with tests and docs extended.

PR #581, opened 2025-08-13
Change how pkexec resolve the canonical location

tht2005:
## Summary
The way pkexec resolve the canonical location
make symlink-based scripts fail to work.

I just change the function used to canonicalize path variable.
realpath: resolve symlink and make path canonical.
g_canonicalize_filename: make path canonical, do not resolve symlink.

For example i have a simple bash script:

$ cat ./a.sh
#!/usr/bin/env bash
echo $0

And i make a soft link: ./b.sh -> ./a.sh

command `pkexec ./b.sh` will print `./a.sh`,
but it should print `./b.sh`.

I make this change because symlink-based scripts
need to know the last symlink preceding it (ex.
python-exec2).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.o
+.cache
 .deps
 /*.bak
 /*.gcda
--- a/src/programs/pkexec.c
+++ b/src/programs/pkexec.c
@@ -19,6 +19,7 @@
  * Author: David Zeuthen <davidz@redhat.com>
  */
 
+#include "glib.h"
 #include <string.h>
 #include <stdlib.h>
 #include <sys/types.h>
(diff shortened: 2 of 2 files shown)


jrybar-rh:
Hello,  
with this change, , the user might unnecessarily get locked away from action/rule files that contain the target path. See the function find_action_for_path().  
@tht2005 , this will not get merged as a default, but if you're willing to implement an argument switch, I'll take another review.  
Many thanks.

jrybar-rh:
@tht2005 , I will close this for now. Please create a new pull request if you want to make the proper change.  
Anyway, many thanks for the PR. It widened my perspective.  
Cheers,  
Jan

PR #582, opened 2025-08-14
Fix locale override issue in privilege escalation

mhduiy:
Remove the pam_env.so configuration that forces system default locale from /etc/default/locale. This prevents user locale preferences from being overridden during privilege escalation.

Previously, if the system default was English but a user had configured Chinese locale, privilege-escalated programs would incorrectly display in English instead of respecting the user's Chinese preference.

## Summary

Remove the pam_env.so configuration that forces system default locale from /etc/default/locale.

If the default language of the system is en_US

1. Set up other languages (such as Japanese)
2. Start gparted (or use pkexec env)
3. View the language displayed on the interface

The interface will display English, but the user chooses Japanese at this time, which is not in line with expectations.


vmihalko (approved):
* Previously the Debian-specific PAM config explicitly pulls system defaults from `/etc/default/locale`. E.g. if system default is en_US and the user’s session is ja_JP, an elevated app started with pkexec may show English due to the PAM-injected system locale.
* The proposed fix (remove the pam_env entry that reads `/etc/default/locale`) is correct and aligns pkexec with the user’s current session locale. Escalated commands inherit the user’s locale preserved by pkexec. The UI and messages match the user’s current session language.
* Scope is Debian-only builds (where `os_type == 'debian'` installs polkit-1.debian).

vmihalko:
You don't need to worry about that; we'll handle it. Thanks for the PR.

PR #596, opened 2025-10-01
DBusMock: CI broken after stop_dbus() removal

jrybar-rh:
Because stop_dbus() was removed from DBusMock implementation, nothing actually stops the system bus in testing environment in the wrapper.py child process, hence the test times out into failure. Adding simple kill(<dbus_PID>) in atexit() is sufficient, but the original implementation in DBusMock seemed more graceful, so it was pulled back in.

## Summary

### DIFF
--- a/test/wrapper.py
+++ b/test/wrapper.py
@@ -5,6 +5,8 @@ import atexit
 import os
 import subprocess
 import sys
+import signal
+import time
 
 import dbus
 import dbus.mainloop.glib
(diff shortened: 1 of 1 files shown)


test/wrapper.py
```diff
@@ -36,6 +38,36 @@ def setup_test_namespace(data_dir):
...
+    """
+    signal.signal(signal.SIGTERM, signal.SIG_IGN)
+    for _ in range(50):
+        try:
+            os.kill(pid, signal.SIGTERM)
+            os.waitpid(pid, os.WNOHANG)
+        except ChildProcessError:
+            break
+        except OSError as e:
+            if e.errno == errno.ESRCH:
```
thesamesam, line 57: This needs an `import errno`.

PR #601, opened 2025-10-14
data: Install an XML catalog to define the policyconfig DTD in

pwithnall:
This allows tools like `xmllint` to look up the installed copy of the
DTD using its public ID, which allows offline validation of policyconfig
files in third-party projects without having to have network access to
download the DTD from
http://www.freedesktop.org/software/polkit/policyconfig-1.dtd.

For example:
```
XML_CATALOG_FILES=/usr/share/polkit-1/catalog.xml xmllint \
  --nonet --noblanks --noout --valid ./my-project.policy
```

---

The use case here is allowing third party projects which ship `.policy` files to validate those files using `xmllint`. `xmllint` needs a DTD catalog file in order to look up the DTD from its public ID (`-//freedesktop//DTD polkit Policy Configuration 1.0//EN`). D-Bus does [similar](https://gitlab.freedesktop.org/dbus/dbus/-/blob/main/doc/catalog.xml.in?ref_type=heads).

Subsequent commits move the XML file installation directory to the more standard `/usr/share/xml/polkit-1`, and add a unit test for the policy file shipped by polkit.

Note that this means the installed DTD has moved from `/usr/share/polkit-1/policyconfig-1.dtd` to `/usr/share/xml/polkit-1/policyconfig-1.dtd`. I can add a compatbility symlink from the old location if you wish.

Signed-off-by: Philip Withnall <pwithnall@gnome.org>

### DIFF
--- a/actions/meson.build
+++ b/actions/meson.build
@@ -1,6 +1,6 @@
 policy = 'org.freedesktop.policykit.policy'
 
-i18n.merge_file(
+policy_file = i18n.merge_file(
   input: policy + '.in',
   output: '@BASENAME@',
   po_dir: po_dir,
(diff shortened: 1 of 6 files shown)


pwithnall:
Pushed some more commits to validate the XML for the D-Bus interface definitions too. This unfortunately means I ended up dropping the EggDBus documentation annotations from the files, as they were not valid XML.

I’m up for implementing alternative solutions to the invalid EggDBus annotations (such as porting the documentation in the interface files from EggDBus to inline XML comments, for example), but I put this together as a first draft for discussion before putting lots of effort into something which might be rejected.

PR #605, opened 2025-10-20
Clean up runtime glib warnings

bluca:
Fix various glib runtime warnings so that polkitd runs clean

### DIFF
--- a/src/polkitbackend/polkitbackendactionpool.c
+++ b/src/polkitbackend/polkitbackendactionpool.c
@@ -576,7 +576,10 @@ ensure_all_files (PolkitBackendActionPool *pool)
                                             &error);
     if (error != NULL)
     {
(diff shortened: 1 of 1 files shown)


src/polkit/polkitunixprocess.c
```diff
@@ -739,7 +739,8 @@ polkit_unix_process_get_pid (PolkitUnixProcess *process)
       if (pid > 0)
         return pid;
 
-      g_error_free (error);
```
vmihalko: This is correct, but we should apply the same logic to the function below: `polkit_unix_process_get_ppid` There is also an unguarded `g_error_free (error);` (line 772).

vmihalko (changes requested):
Thanks for the PR - everything looks correct. I just have one requested change and one consideration.

PR #610, opened 2025-11-10
data: Add licensing and copyright data to D-Bus interface XML files

pwithnall:
These files are likely to be copied into other projects (for example, so proxy objects can be generated from them, or mock interfaces built), so it would be good if their copyright and licensing data went with them.

Copyright data extracted manually by looking at `git log` lists for the two files.

### DIFF
--- a/data/org.freedesktop.PolicyKit1.AuthenticationAgent.xml
+++ b/data/org.freedesktop.PolicyKit1.AuthenticationAgent.xml
@@ -1,3 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Copyright 2009, 2015 Red Hat
+
+  SPDX-License-Identifier: LGPL-2.1-or-later
+
+  This library is free software; you can redistribute it and/or
+  modify it under the terms of the GNU Lesser General Public
+  License as published by the Free Software Foundation; either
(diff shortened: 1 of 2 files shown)


data/org.freedesktop.PolicyKit1.AuthenticationAgent.xml
```diff
@@ -1,3 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Copyright 2009, 2015 Red Hat
+
+  SPDX-License-Identifier: LGPL-2.1-or-later
```
bluca: The SPDX line should be a separate comment, on a single line (see 'Format'): https://spdx.dev/learn/handling-license-info/#how

pwithnall: I think that might be an artifact of the fact the guidance on spdx.dev is using single line comment delimiters. The [guidance on reuse.software](https://reuse.software/tutorial/) shows the SPDX lines can be incorporated into the wider header comment, `reuse lint` is happy with this format, and it’s the way I (and others) have been doing it across multiple projects for a while now!

bluca: Ok fair enough

PR #612, opened 2025-11-18
auth_keep: Disable via sudoers timestamp_timeout=0

danielzgtg:
## Summary

Temporarily inherit sudoers timestamp_timeout=0 upholding security. Permanent'll be #560.

This is a temporary solution to disable auth_keep after #533 broadened it. People upgrading might not realize they cannot walk away from their computer anymore when it has an unlocked non-root shell that just performed some action. Looking at `/usr/share/polkit-1/actions`, worrying ones include PackageKit and UDisks2.

This reads `/etc/sudoers` because the most-affected people would've had edited it. Ideally this can land before the December release. The longer-term task for a permanent solution with nonzero timeouts is #560.

### Draft testing

Draftified while debugging until my Ubuntu VM tty can reproduce the original auth_keep without my changes.


danielzgtg:
> already open and waiting

Nice! I'd prefer that that one get merged soon.

I made this PR in a rush before December in case that one gets delayed.

danielzgtg:
Testing showed infeasibility from `/usr/lib/systemd/system/polkit.service`'s `User=polkitd` being unable to read `/etc/sudoers`.

PR #629, opened 2026-01-04
Introduce FreeBSD CI and make tests pass

arrowd:
See individual commit descriptions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,3 +70,45 @@ jobs:
 
       - name: Build & test
         run: .github/workflows/ci.sh ${{ matrix.phase }} elogind
+
+  freebsd:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v6
+
+      - name: Build & test
+        uses: cross-platform-actions/action@v0.31.0
+        with:
+          operating_system: 'freebsd'
+          version: '15.0'
+          architecture: 'x86_64'
+          run: |
+            # Use latest package set
+            sudo mkdir -p /usr/local/etc/pkg/repos/
+            sudo cp /etc/pkg/FreeBSD.conf /usr/local/etc/pkg/repos/FreeBSD.conf
+            sudo sed -i.bak -e 's|/quarterly|/latest|' /usr/local/etc/pkg/repos/FreeBSD.conf
+
+            echo "::group::Dependencies installation"
+            sudo -E pkg install -y \
+                meson gettext-tools gobject-introspection glib dbus pkgconf \
+                expat duktape-lib consolekit2 py311-dbus py311-python-dbusmock \
+                gtk-doc python3 perl5
+            echo "::endgroup::"
+
+            # gtk-doc fails when calling xsltproc with an obscure error
+            # failed to load external entity "http://www.oasis-open.org/..."
+            # Trick meson in believing that it works fine
+            sudo ln -fs /usr/bin/true /usr/local/bin/gtkdoc-mkhtml
+            sudo ln -fs /usr/bin/true /usr/local/bin/gtkdoc-fixxref
+
+            echo "::group::Build and test"
+            sudo service dbus onestart
+            sudo env ALLOW_SYSTEM_AFFECTING_TESTS=1 \
+                .github/workflows/ci.sh CLANG ConsoleKit
+            echo "::endgroup::"
+
+            # We carried the build as root, so some policy files are now
+            # unreadable by the regular user. The action tries to rsync build
+            # artifacts out of VM, which fails unless we change files owner.
+            sudo chown -R $USER .
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,3 +70,45 @@ jobs:
 
       - name: Build & test
         run: .github/workflows/ci.sh ${{ matrix.phase }} elogind
+
+  freebsd:
+    runs-on: ubuntu-24.04
+    concurrency:
+      group: freebsd-${{ github.workflow }}-${{ github.ref }}-${{ matrix.phase }}
+      cancel-in-progress: true
+    defaults:
+      run:
+        shell: cpa.sh {0}
+
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v6
+
+      - name: Build & test
+        uses: cross-platform-actions/action@5ea7e8e4677bd726033a10b094ba1c5762b15dee # v.1.3.0
+        with:
+          operating_system: 'freebsd'
+          version: '15.1'
+          architecture: 'x86_64'
+          run: |
+            # Use latest package set
+            sudo mkdir -p /usr/local/etc/pkg/repos/
+            sudo cp /etc/pkg/FreeBSD.conf /usr/local/etc/pkg/repos/FreeBSD.conf
+            sudo sed -i.bak -e 's|/quarterly|/latest|' /usr/local/etc/pkg/repos/FreeBSD.conf
+
+            echo "::group::Dependencies installation"
+            sudo -E pkg install -y \
+                meson gettext-tools gobject-introspection glib dbus pkgconf \
+                expat duktape-lib consolekit2 py312-dbus py312-python-dbusmock \
+                gtk-doc python3 perl5
+            echo "::endgroup::"
+
+            echo "::group::Build and test"
+            sudo service dbus onestart
+            sudo .github/workflows/ci.sh CLANG ConsoleKit
+            echo "::endgroup::"
+
+            # We carried the build as root, so some policy files are now
+            # unreadable by the regular user. The action tries to rsync build
+            # artifacts out of VM, which fails unless we change files owner.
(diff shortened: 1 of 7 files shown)


arrowd:
Ping.

arrowd:
Ping again.

arrowd:
Monthly ping.

test/wrapper.py
```diff
@@ -34,15 +36,11 @@ def setup_test_namespace(data_dir):
...
     except AttributeError:
         print("Python 3.12 is required for os.unshare(), skipping")
-        sys.exit(77)
-    except OSError as e:
-        if e.errno == errno.EINVAL:
-            print("os.unshare(os.CLONE_NEWNS|os.CLONE_NEWUSER) not supported, skipping")
-            sys.exit(77)
-        raise
+        return False
+    return True
```
jrybar-rh, line 48: Removing OSError results in unhandled exception. False can be returned, but the exception needs to be caught first.

test/wrapper.py
```diff
@@ -74,6 +72,43 @@ def stop_dbus(pid: int) -> None:
     signal.signal(signal.SIGTERM, signal.SIG_DFL)
 
 
+def setup_test_unconstrained_linux(data_dir):
+    shutil.copytree(os.path.join(data_dir, "etc"), "/etc", dirs_exist_ok=True)
```
jrybar-rh, line 81: If I'm not mistaken, this fn is called when os.unshare() fails, so this overwrites live /etc on host. It cannot be presumed that the test is always run in a scrap vm/container or setup_test_namespace() doesn't fail.

test/wrapper.py
```diff
@@ -74,6 +72,43 @@ def stop_dbus(pid: int) -> None:
...
+def setup_test_unconstrained_freebsd(data_dir):
+    # The /etc/passwd file on FreeBSD has a slightly different structure,
+    # so instead of copying it we parse the source file and invoke 'pw useradd'
+    passwd = os.path.join(data_dir, "etc", "passwd")
+    with open(passwd, 'r') as file:
+        for line in file:
+            fields = line.split(":")
+            if fields[0] == "root":
+                continue
+            subprocess.call(["pw", "useradd", "-n", fields[0], "-u", fields[2], "-c", fields[4], "-d", fields[5], "-s", fields[6].strip()])
```
jrybar-rh: The number of fields should be checked for first if no return code is handled.

.github/workflows/ci.yml
```diff
@@ -70,3 +70,45 @@ jobs:
...
         run: .github/workflows/ci.sh ${{ matrix.phase }} elogind
+
+  freebsd:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v6
+
+      - name: Build & test
+        uses: cross-platform-actions/action@v0.31.0
```
jrybar-rh: Hash might be more helpful than version (cross-platform-actions doesn't have public members. The more security, the better.)

test/wrapper.py
arrowd: The `_unconstrained_*` variants of the test are only invoked if the user provided the `ALLOW_SYSTEM_AFFECTING_TESTS=1` env var. Maybe we should rename it to `ALLOW_SYSTEM_DESTRUCTIVE_TESTS` to underline the fact, that they should only be executed in a throw-away environments.

.github/workflows/ci.yml
```diff
@@ -70,3 +70,49 @@ jobs:
...
+
+            # gtk-doc fails when calling xsltproc with an obscure error
+            # failed to load external entity "http://www.oasis-open.org/..."
+            # Trick meson in believing that it works fine
+            sudo ln -fs /usr/bin/true /usr/local/bin/gtkdoc-mkhtml
+            sudo ln -fs /usr/bin/true /usr/local/bin/gtkdoc-fixxref
+
+            echo "::group::Build and test"
+            sudo service dbus onestart
+            sudo env ALLOW_SYSTEM_AFFECTING_TESTS=1 \
```
mrc0mmand: I'd move the `ALLOW_SYSTEM_AFFECTING_TESTS=1` variable definition to `.github/workflows/ci.sh` and export it there, since we want to run the "system affecting tests" in all GHA CIs.

.github/workflows/ci.yml
```diff
@@ -70,3 +70,49 @@ jobs:
...
+            sudo sed -i.bak -e 's|/quarterly|/latest|' /usr/local/etc/pkg/repos/FreeBSD.conf
+
+            echo "::group::Dependencies installation"
+            sudo -E pkg install -y \
+                meson gettext-tools gobject-introspection glib dbus pkgconf \
+                expat duktape-lib consolekit2 py311-dbus py311-python-dbusmock \
+                gtk-doc python3 perl5
+            echo "::endgroup::"
+
+            # gtk-doc fails when calling xsltproc with an obscure error
```
mrc0mmand: This is strange, I can't reproduce anything like this locally on FreeBSD 15.1 - do you have any logs where this happened?

test/wrapper.py
```diff
@@ -74,6 +77,45 @@ def stop_dbus(pid: int) -> None:
...
+            members = fields[3].strip()
+            if len(members) > 0:
+                cmd.append("-M")
+                cmd.append(members)
+            if fields[0] == "root":
+                cmd = ["pw", "groupmod", "-g", "0", "-l", "root"]
+            subprocess.call(cmd)
+    # netgroup can be used as is
+    shutil.copy2(os.path.join(data_dir, "etc", "netgroup"), "/etc/netgroup")
+    # 10-testing.rules calls into /bin/true, but on FreeBSD it is located
```
mrc0mmand: Feel free to update `10-testing.rules` to use `/usr/bin/true` instead, so you can drop this.

test/wrapper.py
```diff
@@ -74,6 +77,45 @@ def stop_dbus(pid: int) -> None:
     signal.signal(signal.SIGTERM, signal.SIG_DFL)
 
 
+def setup_test_unconstrained_linux(data_dir):
+    shutil.copytree(os.path.join(data_dir, "etc"), "/etc", dirs_exist_ok=True)
+
+def setup_test_unconstrained_freebsd(data_dir):
+    # The /etc/passwd file on FreeBSD has a slightly different structure,
```
mrc0mmand: I'm not really a fan of this part. I'd just create the `passwd` and `group` files in `test/data/etc/` with `.bsd` suffix (or something like that) in the expected format, and use them here to replace the original `passwd` and `group`.

test/data/etc/polkit-1/rules.d/10-testing.rules
```diff
@@ -140,7 +140,7 @@ polkit.addRule(function(action, subject) {
 polkit.addRule(function(action, subject) {
     if (action.id == "net.company.spawning.successful_helper") {
         try {
-            polkit.spawn(["/bin/true"]);
+            polkit.spawn(["/usr/bin/true"]);
```
mrc0mmand: *sigh*, looks like this time it's Alpine that's misbehaving, because:

```
/ # which true
/bin/true
/ # ls -l /bin/true
lrwxrwxrwx    1 root     root            12 Feb 13  2025 /bin/true -> /bin/busybox
```

However, given that `polkit.spawn()` calls `g_spawn_async_with_pipes()` with `G_SPAWN_SEARCH_PATH`, the absolute path shouldn't be necessary here. So let's just use `true` (and `false` below), and hope for the best.

mrc0mmand (commented):
Also, please squash the last two commits, since the latter one removes something the former commit introduces.

mrc0mmand (approved):
LGTM, thanks.

The remaining CI fail is unrelated - it's a bug in Alpine's `fortify-headers` package, which has been already resolved in upstream https://github.com/jvoisin/fortify-headers/commit/b9121e4d679b5fccb69abcb58b05a6be5c7191dc), but hasn't been pulled into Alpine yet.

(EDIT: it actually has been pulled into Alpine, but just ~three hours ago ATTOW - https://gitlab.alpinelinux.org/alpine/aports/-/commit/6f16884e9f49f56df3f41c1ab07f221e4a0ae051)

.github/workflows/ci.yml
```diff
@@ -70,3 +70,42 @@ jobs:
 
       - name: Build & test
         run: .github/workflows/ci.sh ${{ matrix.phase }} elogind
+
+  freebsd:
+    runs-on: ubuntu-24.04
```
mrc0mmand, line 75: Ah, one more thing that I overlooked: please add a concurrency section here, so the freebsd CI job is cancelled on force-push when a new one is started. I.e. something like:

```yaml
    concurrency:
      group: freebsd-${{ github.workflow }}-${{ github.ref }}-${{ matrix.phase }}
      cancel-in-progress: true
```

Similarly how it's done for the already existing jobs.

test/wrapper.py
jrybar-rh: This sounds better, more alerting. Can you please push it? Everything else LGTM.

jrybar-rh (approved):

PR #631, opened 2026-01-06
Support locked system accounts with systemd 257+

stoeckmann:
## Summary
Beginning with version 257, systemd supports locked system accounts. Use them if available.

A locked system user is more restrictive than a system user with a locked password. The former disallows any form of logins, while the latter could be accessed through other password-less means.

This is just another defensive step, i.e. no security bugfix!

See also https://archlinux.org/todo/change-sysusers-to-fully-locked-system-accounts/

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -1,6 +1,7 @@
 service_conf = {
   'libprivdir': pk_prefix / pk_libprivdir,
   'polkitd_user': polkitd_user,
+  'polkitd_user_lock': polkitd_user_lock,
   'polkitd_uid': polkitd_uid,
 }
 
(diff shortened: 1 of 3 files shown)


bluca:
I'm not sure it's worth having an option for this? Just set the `!` always? If some distribution backports new polkit to an older version (systemd v256 is from 2024), they can provide their own sysusers.d file, as it's trivial to do so, while a meson option has to be maintained forever

stoeckmann:
> I'm not sure it's worth having an option for this?

It wouldn't be, but this is not an option. :)

The version is retrieved anyway and this PR just adds a check of this version.

> Just set the `!` always?

I could live with that as well.

jrybar-rh:
> Just set the `!` always? If some distribution backports new polkit to an older version (systemd v256 is from 2024), they can provide their own sysusers.d file

Yeah, true. Especially when it is recommended to always have the lock there, the distros that do not reflect that can always deploy their own sysusers file.

@stoeckmann, will you be able to push the change, or should I close this and make a fast change myself?
Anyway, thank you.

stoeckmann:
Hi, I've just updated the PR

bluca (approved):

PR #645, opened 2026-03-09
fix race in deleting agent after unregistering bus

JakubSzczudlo:
## Summary
Fix race in deleting agent after unregistering bus
Sometimes function that get agent is slower than operating system and pidfd is deleted faster that it can use it to find right agent in hash map.

Make function calculating hash for subject stable so agent can be deleted after process exited. Add agent correctness function that check if subject pid doesn't change. Use subject initial pid in functions for checking equal and calculating hash.

Can be reproduced by printing all Gobjects in loop in polkitd and using : 
`while true; do   for i in $(seq 1 10000); do loginctl enable-linger $USER; sleep 0.1; done; echo; done`
If loop run for more than few hours memory usage increase can be also seen in top command

Related to #150

### DIFF
--- a/src/polkit/polkitsystembusname.c
+++ b/src/polkit/polkitsystembusname.c
@@ -271,6 +271,12 @@ polkit_system_bus_name_exists_sync (PolkitSubject   *subject,
   return ret;
 }
 
+static gboolean
+polkit_system_bus_name_check (PolkitSubject   *subject)
+{
+  return TRUE;
+}
+
 static void
 exists_in_thread_func (GSimpleAsyncResult *res,
                        GObject            *object,
@@ -338,6 +344,7 @@ subject_iface_init (PolkitSubjectIface *subject_iface)
   subject_iface->exists        = polkit_system_bus_name_exists;
   subject_iface->exists_finish = polkit_system_bus_name_exists_finish;
   subject_iface->exists_sync   = polkit_system_bus_name_exists_sync;
+  subject_iface->check         = polkit_system_bus_name_check;
 }
 
 /* ---------------------------------------------------------------------------------------------------- */
--- a/src/polkit/polkitunixprocess.c
+++ b/src/polkit/polkitunixprocess.c
@@ -390,7 +390,6 @@ polkit_unix_process_constructed (GObject *object)
       if (pidfd >= 0)
         {
           process->pidfd = pidfd;
-          process->pid = 0;
         }
     }
 
@@ -1060,8 +1059,9 @@ static guint
 polkit_unix_process_hash (PolkitSubject *subject)
 {
   PolkitUnixProcess *process = POLKIT_UNIX_PROCESS (subject);
+  guint hash_value = process->pid + process->start_time;
 
-  return g_direct_hash (GSIZE_TO_POINTER ((polkit_unix_process_get_pid(process) + process->start_time))) ;
+  return g_direct_hash (GSIZE_TO_POINTER (hash_value)) ;
 }
 
 static gboolean
@@ -1076,8 +1076,8 @@ polkit_unix_process_equal (PolkitSubject *a,
   process_a = POLKIT_UNIX_PROCESS (a);
   process_b = POLKIT_UNIX_PROCESS (b);
 
-  pid_a = polkit_unix_process_get_pid(process_a);
-  pid_b = polkit_unix_process_get_pid(process_b);
+  pid_a = process_a->pid;
+  pid_b = process_b->pid;
 
   pidfd_a = polkit_unix_process_get_pidfd(process_a);
   pidfd_b = polkit_unix_process_get_pidfd(process_b);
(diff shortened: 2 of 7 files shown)


JakubSzczudlo:
@bluca maybe you can check that changes?

jrybar-rh:
This PR is in the queue for review. The strategy of precedence depends (among other things) on the amount of changes made to the codebase. Please be patient.

JakubSzczudlo:
I understand. Thanks for answer @jrybar-rh.

src/polkit/polkitunixprocess.c
```diff
@@ -390,7 +390,6 @@ polkit_unix_process_constructed (GObject *object)
       if (pidfd >= 0)
         {
           process->pidfd = pidfd;
-          process->pid = 0;
```
jrybar-rh, line 393: No, this reintroduces the PID-reuse vulnerability, and neuters the entire idea of using PIDFDs instead of PIDs.

src/polkit/polkitunixprocess.c
```diff
@@ -1142,6 +1142,14 @@ polkit_unix_process_exists_sync (PolkitSubject   *subject,
   return ret;
 }
 
+static gboolean
+polkit_unix_process_check (PolkitSubject *subject)
```
jrybar-rh, line 1146: I haven't thought of it much, but maybe polkit_unix_process_exists(), which is right under this function, could be used for this conveniently.

src/polkitbackend/polkitbackendinteractiveauthority.c
```diff
@@ -1785,6 +1787,16 @@ authentication_agent_unref (AuthenticationAgent *agent)
     }
 }
 
+static void
+check_agent_subject (AuthenticationAgent *agent, PolkitSubject *subject)
+{
+  gboolean is_subject_correct = polkit_subject_state_check(subject);
+  if (is_subject_correct == FALSE)
+  {
+    agent = NULL;
```
jrybar-rh, line 1796: This does nothing. Only the local variable is NULLified. To make it have effect, the function parameter must be `AuthenticationAgent **agent`, the function must be called with `&agent` and then `*agent=NULL` would work.

src/polkitbackend/polkitbackendinteractiveauthority.c
```diff
@@ -1996,10 +2017,12 @@ get_authentication_sessions_initiated_by_system_bus_unique_name (PolkitBackendIn
...
 
   g_hash_table_iter_init (&hash_iter, priv->hash_scope_to_authentication_agent);
-  while (g_hash_table_iter_next (&hash_iter, NULL, (gpointer) &agent))
+  while (g_hash_table_iter_next (&hash_iter, (gpointer) &subject, (gpointer) &agent))
     {
       GList *l;
 
+      check_agent_subject(agent, subject);
+
       for (l = agent->active_sessions; l != NULL; l = l->next)
```
jrybar-rh, line 2026: ditto

src/polkitbackend/polkitbackendinteractiveauthority.c
```diff
@@ -1958,6 +1975,9 @@ get_authentication_session_for_uid_and_cookie (PolkitBackendInteractiveAuthority
        * who is using our own setuid helper will automatically be updated
        * to the new API.
        */
+      
+      check_agent_subject(agent, subject);
+      
       if (uid != (uid_t)-1)
         {
           if (agent->creator_uid != uid)
```
jrybar-rh, line 1983: If `check_agent_subject()` worked, this would crash immediately, as NULL would be approached for `->creator_uid`, leading to NULL ptr exception. Just as well as in the iterations below.

src/polkitbackend/polkitbackendinteractiveauthority.c
jrybar-rh, line 1987: ditto

src/polkit/polkitsystembusname.c
```diff
@@ -271,6 +271,12 @@ polkit_system_bus_name_exists_sync (PolkitSubject   *subject,
   return ret;
 }
 
+static gboolean
+polkit_system_bus_name_check (PolkitSubject   *subject)
+{
+  return TRUE;
```
jrybar-rh, line 277: Generally, adding a dummy method to every PolkitSubject type except UnixProcess is step back, IMO. I realize and appreciate what is tried to fix here, but I'd really rather go the `polkit_unix_process_exists_sync()` way. There already are functions in the codebase that do the same thing.

jrybar-rh (commented):
Thank you for tackling this bug, I appreciate the effort, yet I cannot but close this PR as a different approach on architectural design is needed and the current implementation contains many serious issues. There are few nitpicks about naming convention, unnecessary (and misalligned) indentation changes. Instead of reiterating heavily on this PR, I'd suggest to open a new one which provides a cleaner foundation to build on. Looking forward to the next review!

PR #646, opened 2026-03-27
fix endless memory ready using getline

msmeissn:
See CVE-2026-4897

As this is run setuid root, it could bypass user based limits and run the system out of memory.

## Summary

getline can be passed a buffer, this avoids attackers passing in gigabytes of garbage

https://bugzilla.redhat.com/show_bug.cgi?id=2451739

# No newline: getline() keeps reallocating until OOM or ENOMEM
python3 -c "print('A'*200000000, end='')" | /usr/lib/polkit-1/polkit-agent-helper-1 $(whoami)

### DIFF
--- a/src/polkitagent/polkitagenthelperprivate.c
+++ b/src/polkitagent/polkitagenthelperprivate.c
@@ -48,6 +48,7 @@ _polkit_clearenv (void)
 char *
 read_cookie (int argc, char **argv)
 {
+  char buf[200];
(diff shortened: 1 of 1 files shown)


jrybar-rh:
Hello, thank you for your PR.  
Since Aisle.com reported the issue and provided the fix before I even requested for the CVE number, I guess it is fair to merge their fix.  
Anyway, you're quick and vigilant, thanks. I didn't notice the process of disclosing the security report went so fast.

PR #648, opened 2026-04-07
polkitd: Add sigterm_id to store the SIGTERM event source ID

wangrong1069:
The SIGINT source ID was incorrectly overwritten by the SIGTERM handler registration, causing the SIGINT source to never be removed during shutdown. This led to a resource leak.

Add separate sigterm_id variable to properly track and cleanup both signal sources.

### DIFF
--- a/src/polkitbackend/polkitd.c
+++ b/src/polkitbackend/polkitd.c
@@ -308,6 +308,7 @@ main (int    argc,
   GOptionContext *opt_context;
   guint name_owner_id;
   guint sigint_id;
+  guint sigterm_id;
   guint sighup_id;
   guint expiration_seconds;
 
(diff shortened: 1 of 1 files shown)


vmihalko (dismissed):
[EDIT] Sorry for the noise - an earlier approval on this PR was posted in error (automated assistant). I will post my review ASAP.

vmihalko (approved):
The PR fixes a copy-paste mistake: SIGTERM was registered correctly but its source ID was stored in sigint_id, leaking the SIGINT source and only partially cleaning up on exit. The patch restores symmetric handling for all three signals (SIGINT, SIGTERM, SIGHUP).

Thank you for contributing!

PR #656, opened 2026-05-03
Avoid periodic polling for non-existent polkit rules directories

Pablito2020:
## Summary

Currently, `polkitbackendduktapeauthority.c` watches a set of hardcoded rule directories. If some of these directories do not exist, GLib’s inotify backend falls back to polling: every 4 seconds it attempts to add the inotify watch again.

This can result in repeated filesystem lookups for paths that are known to be unused on a given system. I noticed this while experimenting with eBPF and tracing `__lookup_slow`:

```bash
sudo bpftrace -e 'kprobe:__lookup_slow { printf("Name: %s, Parent: %p, Flags: %d\n", str(((struct qstr *)arg0)->name), arg1, arg2); }'
```

On systems where polkit rules are stored only in specific locations (and we are a 100% sure the user can't add more rules, since they're declared and immutable, like for example NixOS + impermanence setup), repeatedly checking missing directories is unnecessary. Even if the underlying filesystem is usually something lightweight like `tmpfs`, avoiding this polling is still preferable when we already know that certain rule directories will never be used.

This PR introduces an optional environment variable, empty by default, that allows users or distributions to explicitly specify the directories where polkit should look for rules. When set, `polkitd` will use those directories instead of watching all the hardcoded rule paths. This allows systems with known rule locations to avoid the repeated polling caused by missing directories.

The polling behavior comes from GLib's inotify fallback implementation. When `inotify_add_watch()` fails, GLib adds the path to a "missing" watch list and periodically retries the watch every 4 seconds by default, some of the interesting links of the glib source code are the following:

- [(glib) set ups a callback for executing `im_scan_missing` each SCAN_MISSING_TIME (4 secs by default)](https://github.com/GNOME/glib/blob/main/gio/inotify/inotify-missing.c#L60)
- [(glib) `im_scan_missing`, tries to inotify the dirs, if it fails because they don't exist, don't remove the callback](https://github.com/GNOME/glib/blob/main/gio/inotify/inotify-missing.c#L108)

This PR may also be related to issue #90, although that issue mentions both configuration and rules directories, while this change only affects rule directories.

I am opening this as a draft because I have not yet checked whether there has already been previous discussion or an existing proposal around this idea, or whether the project would consider this approach viable.

I am also still pretty new to this codebase (first time with gnome glib...), and I used some AI assistance to help me get familiar with it, so please feel free to be strict in the review. :)

You can reproduce the "issue" (if it can be considered an issue?) via the eBPF command that I told you in the summary section. Be aware that, if you have all the "hardcoded" folders created, you may have to delete one of them to see the polling behavior.

### DIFF
--- a/docs/man/polkitd.xml
+++ b/docs/man/polkitd.xml
@@ -69,7 +69,7 @@
           <para>
             If set, overrides the default directories used to load JavaScript
             authorization rules. The value is a colon-separated list of
-            directories.
+            directories. Empty path elements are ignored.
           </para>
         </listitem>
       </varlistentry>
--- a/docs/man/polkitd.xml
+++ b/docs/man/polkitd.xml
@@ -61,6 +61,21 @@
     </para>
   </refsect1>
 
+  <refsect1 id="polkitd-environment"><title>ENVIRONMENT</title>
+    <variablelist>
+      <varlistentry>
+        <term><envar>POLKIT_RULES_DIRS</envar></term>
+        <listitem>
+          <para>
+            If set, overrides the default directories used to load JavaScript
+            authorization rules. The value is a colon-separated list of
+            directories.
+          </para>
+        </listitem>
+      </varlistentry>
+    </variablelist>
+  </refsect1>
+
   <refsect1 id="polkitd-author"><title>AUTHOR</title>
     <para>
       Written by David Zeuthen <email>davidz@redhat.com</email> with
--- a/src/polkitbackend/polkitbackendduktapeauthority.c
+++ b/src/polkitbackend/polkitbackendduktapeauthority.c
@@ -30,6 +30,8 @@
 /* Built source and not too big to worry about deduplication */
 #include "initjs.h" /* init.js */
 
+#define POLKIT_RULES_DIRS_ENV_VAR "POLKIT_RULES_DIRS"
+
 /**
  * SECTION:polkitbackendjsauthority
  * @title: PolkitBackendJsAuthority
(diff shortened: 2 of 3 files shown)


bluca:
Adding an environment variable to hack around the issue doesn't seem like the best solution. It should be possible to instead do some sort of exponential backoff in the retries (up to a certain cap), so that the directory probing becomes less frequent, say cap it a once every 10 minutes

Pablito2020:
Ok, that makes sense to me.

I don't think GLib currently exposes a way to configure this kind of behavior. My question is whether this feature should be proposed upstream in GLib.

I could imagine an interface like this:
```c
void
g_file_monitor_set_missing_watch_retry (GFileMonitor *monitor,
                                        guint         initial_msecs, // initial retry timeout
                                        guint         max_msecs, // max retry timeout (eg: 10 minutes)
                                        gdouble       multiplier // the exponential multiplier until arriving at max_msecs
);
```
This would adjust the internal retry parameters of struct inotify_sub in the inotify backend we are using. It might also be applicable to the libinotify-kqueue backend, although I’m not completely sure about that. For other implementations, it could simply be a no-op.

Alternatively, the function could return a status code so callers can check whether the backend supports this configuration and whether the update was applied successfully.

The main concern I see with this approach is that it would publicly expose implementation details of GFileMonitor that only apply to a subset of the available backends... Another option is to add a new flag on [FileMonitorFlags](https://docs.gtk.org/gio/flags.FileMonitorFlags.html) that enable us to fail if the file/directory does not exist (without trying polling), for letting the user of the library decide if they want to retry and how.

So, what do you think would be the best approach here?

PR #658, opened 2026-05-05
polkitbackend: fix synthetic-caller leak and NULL unref in push_subject

Rackme:
## Summary

Two small memory-management fixes for `polkitd`, uncovered while auditing the per-`CheckAuthorization` hot path. Combined diff: 7 insertions, 2 deletions across 2 files. Both bugs are pre-existing — not regressions from the recent leak fixes (#640 / #653 / #654 / #571 / #576 / #563).

1. **`polkit_backend_interactive_authority_check_authorization()`** leaks one `PolkitSystemBusName` per self-invocation (calls where `caller == NULL`). The "called from ourselves" branch allocates a fresh caller via `polkit_system_bus_name_new()` and never unrefs it.
2. **`push_subject()`** calls `g_object_unref(NULL)` when `polkit_system_bus_name_get_process_sync()` fails (bus name vanishes mid-query), producing a `GLib-GObject-CRITICAL` warning in the journal. Not a leak per se, but matches the cleanup style of the recently-merged GError fixes.

### Patch 1 — `polkitbackend: don't leak the synthetic caller used for self-invocations`

When `polkit_backend_interactive_authority_check_authorization()` is invoked with `caller == NULL` (the "called from ourselves" hack near the top of the function), it allocates a fresh `PolkitSystemBusName` via `polkit_system_bus_name_new()` and stores it in the local `caller` parameter. The function's `out:` cleanup never unrefs it, because in the normal D-Bus path `caller` is owned by the function's caller.

The fix tracks the synthesised caller separately as `local_caller` and unrefs it at `out:`; the normal caller-parameter-borrowed flow is unchanged.

**Verification:** under an ASAN build, exercise the `caller == NULL` branch by calling `polkit_backend_authority_check_authorization()` with `caller=NULL`. Without the patch, ASAN reports a leak rooted at `polkit_system_bus_name_new`. With the patch, clean. The branch isn't exercised by the in-tree code paths, so external triggering is hard — ASAN + a synthetic test is the most reliable verification.

### Patch 2 — `polkitbackend: avoid g_object_unref(NULL) in push_subject error path`

When `polkit_system_bus_name_get_process_sync()` fails (e.g. the bus name vanished mid-query), `push_subject()` `goto`s `out:` with `process` still `NULL`. The `out:` block then calls `g_object_unref(process)` because the subject is still a `SYSTEM_BUS_NAME`, firing `GLib-GObject-CRITICAL ** g_object_unref: assertion 'G_IS_OBJECT (object)' failed` in the journal.

The fix guards the unref with `process != NULL`. Trivially small and matches the cleanup style of the recently merged GError-related fixes.

**Verification:** send a `CheckAuthorization` for a system bus name that vanishes mid-query and watch the journal for the absence of the `G_IS_OBJECT` critical. The race is hard to hit reliably; for practical purposes code review is sufficient.

### DIFF
--- a/src/polkitbackend/polkitbackendduktapeauthority.c
+++ b/src/polkitbackend/polkitbackendduktapeauthority.c
@@ -544,7 +544,7 @@ else
   ret = TRUE;
 
  out:
-  if (POLKIT_IS_SYSTEM_BUS_NAME (subject))
+  if (POLKIT_IS_SYSTEM_BUS_NAME (subject) && process != NULL)
     g_object_unref (process);
   free (session_str);
   free (seat_str);
--- a/src/polkitbackend/polkitbackendinteractiveauthority.c
+++ b/src/polkitbackend/polkitbackendinteractiveauthority.c
@@ -1008,6 +1008,7 @@ polkit_backend_interactive_authority_check_authorization (PolkitBackendAuthority
 {
   PolkitBackendInteractiveAuthority *interactive_authority;
   PolkitBackendInteractiveAuthorityPrivate *priv;
+  PolkitSubject *local_caller = NULL;
   gchar *caller_str;
   gchar *subject_str;
   PolkitIdentity *user_of_caller;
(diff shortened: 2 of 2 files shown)


Rackme:
[polkit-reproduce-kit.zip](https://github.com/user-attachments/files/27474796/polkit-reproduce-kit.zip)

vmihalko (dismissed):
[EDIT] Sorry for the noise - an earlier approval on this PR was posted in error (automated assistant). I will post my review ASAP.

vmihalko (commented):
[EDIT] Sorry for the noise - an earlier APPROVED review on this PR was posted in error (automated assistant) I will deliver the review ASAP

vmihalko (approved):
LGTM. 
First commit correctly separates borrowed vs owned caller (`local_caller` - can’t unref caller unconditionally because the D-Bus layer owns it). 
Second commit correctly guards `g_object_unref(process)` when `get_process_sync()` fails.

Thank you for contributing!

PR #663, opened 2026-05-30
agent helper: Allow access to tpm in agent helper sandbox

RazeLighter777:
## Summary
One line change to polkit-agent-helper service unit to allow access to tpm devices an explicit DeviceAllow entry.

This allows access for pam modules that make use of the tpm to be utilized with polkit.

Allowing access to the tpm (trusted platform module) is useful for authorization and authentication tasks, such as gating polkit authorization on a tpm backed credential or system PCR registers.

Residual risk is primarily DoS via lockout exhaustion attacks and is bounded by the TPM's own auth/DA controls; all other unit hardening is unchanged and DevicePolicy is retained as closed.

See issue https://github.com/RazeLighter777/pinpam/issues/4#issuecomment-3815461955 for an example of an upstream project requiring a workaround for this issue.

### DIFF
--- a/data/polkit-agent-helper@.service.in
+++ b/data/polkit-agent-helper@.service.in
@@ -15,6 +15,7 @@ LockPersonality=yes
 MemoryDenyWriteExecute=yes
 NoNewPrivileges=yes
 PrivateDevices=yes
+DeviceAllow=char-tpm rw
 PrivateNetwork=yes
 PrivateTmp=yes
 ProtectControlGroups=yes


RazeLighter777:
Note: https://github.com/polkit-org/polkit/pull/626
(Closed) is related to this, but this seeks to add a specific device exception rather than changing overall DevicePolicy. Carefully expanding polkit’s agent sandbox seems justified for this case. The primary job of the TPM is to perform authorization and authentication tasks like this.

I’m aware of shipping a systemd service override, and it fact, that is what I do for my package. Shipping a systemd specific workaround isn’t ideal for this; PAM modules shouldn’t be aware of the init systems at all.

I believe narrow exceptions for specific, authorization/authentication related devices are justified. Especially because PAM modules are almost always installed for multiple Pam services, not just Polkit, meaning that the restriction, while well intended and useful, results in unexpected and hard to diagnose behavior when a user sees their module work with sudo and their login manager, but not polkit.

I think tpm devices specifically are a justifiable addition to the policy.

Maryse47:
As esplained in https://github.com/polkit-org/polkit/pull/626#issuecomment-3693173839 without `PrivateDevices=no` this change does nothing:

```
$ systemd-run -P --service-type oneshot -p PrivateDevices=yes ls /dev
Running as unit: run-p29618-i8139.service
char  core  fd  full  hugepages  log  mqueue  null  ptmx  pts  random  shm  stderr  stdin  stdout  tty  urandom  zero

$ systemd-run -P --service-type oneshot -p PrivateDevices=yes -p DeviceAllow="char-tpm r" ls /dev
Running as unit: run-p29517-i15403.service
char  core  fd  full  hugepages  log  mqueue  null  ptmx  pts  random  shm  stderr  stdin  stdout  tty  urandom  zero
```

jrybar-rh:
It is obvious that issues of this type get more and more numerous every week. I accept that, because removing SUID bit from polkit-agent-helper is a milestone that looks like a hurdle for some. Partially it is. Objectively, security should be paramount in polkit as an authorization suite. Some might call that an inconvenience. Even SELinux/Apparmor might look like a nuisance, but your distro certainly got along with it. It requires a bit more energy to make it work seamlessly, but it is worth the security it should provide.  

I don't doubt there might be a process that can straightforward the experience this change brings, maybe even a new system-wide concept, and it only needs time to find this process.  

I thought that it might make sense to leave this to the distro package maintainers, i.e. the drop-in overrides should be parts of the packages that need the override, distro-specific. Making changes in the unit file here on polkit upstream doesn't make sense. The unit file would turn into a list of exceptions for all the projects that use PAM modules and interact with polkit.  

@bluca, does it make sense?

bluca:
> It is obvious that issues of this type get more and more numerous every week. I accept that, because removing SUID bit from polkit-agent-helper is a milestone that looks like a hurdle for some. Partially it is. Objectively, security should be paramount in polkit as an authorization suite. Some might call that an inconvenience. Even SELinux/Apparmor might look like a nuisance, but your distro certainly got along with it. It requires a bit more energy to make it work seamlessly, but it is worth the security it should provide.
> 
> I don't doubt there might be a process that can straightforward the experience this change brings, maybe even a new system-wide concept, and it only needs time to find this process.
> 
> I thought that it might make sense to leave this to the distro package maintainers, i.e. the drop-in overrides should be parts of the packages that need the override, distro-specific. Making changes in the unit file here on polkit upstream doesn't make sense. The unit file would turn into a list of exceptions for all the projects that use PAM modules and interact with polkit.
> 
> @bluca, does it make sense?

Yes this makes perfect sense, it's up to the pam modules to ship the drop in overrides they need. Making everything worse for everyone by default is definitely not the best solution. Shipping a drop in override is trivial and any pam package can do it easily, and they are in the best position to know what access their module actually needs, and doesn't need to wait for a new polkit release every time they change something.

RazeLighter777:
> > It is obvious that issues of this type get more and more numerous every week. I accept that, because removing SUID bit from polkit-agent-helper is a milestone that looks like a hurdle for some. Partially it is. Objectively, security should be paramount in polkit as an authorization suite. Some might call that an inconvenience. Even SELinux/Apparmor might look like a nuisance, but your distro certainly got along with it. It requires a bit more energy to make it work seamlessly, but it is worth the security it should provide.
> 
> > 
> 
> > I don't doubt there might be a process that can straightforward the experience this change brings, maybe even a new system-wide concept, and it only needs time to find this process.
> 
> > 
> 
> > I thought that it might make sense to leave this to the distro package maintainers, i.e. the drop-in overrides should be parts of the packages that need the override, distro-specific. Making changes in the unit file here on polkit upstream doesn't make sense. The unit file would turn into a list of exceptions for all the projects that use PAM modules and interact with polkit.
> 
> > 
> 
> > @bluca, does it make sense?
> 
> 
> 
> Yes this makes perfect sense, it's up to the pam modules to ship the drop in overrides they need. Making everything worse for everyone by default is definitely not the best solution. Shipping a drop in override is trivial and any pam package can do it easily, and they are in the best position to know what access their module actually needs, and doesn't need to wait for a new polkit release every time they change something.

Fair enough. I do understand this; if you had to add exceptions for every hardware device like usb fingerprint readers, smart cards etc it would be hairy and weaken security generally. I do in fact ship an override for my pam module for polkit specifically.

It's ironic because my pam module does sandboxing in itself through Landlock too, I'm in full favor of the sandboxing.

It's just confusing when sudo and login services have no sandboxing and suddenly the module doesn't work specifically on polkit. There was basically no log surface or documentation for this sandbox existing or it denying anything; until someone tried to use the module and I investigated it, it was unclear what was going on.

(I'd be willing to contribute documentation)

Do you have a recommendation for a interface for this than general overrides? I think if everyone is shipping their own overrides, they may conflict or overbroaden the sandbox for everyone. It would be nice to be able to say "my pam module needs abc, this other pam module needs xyz" and to be able to ship overrides that apply specifically to the invokation of that module and not any other.

Otherwise I feel some bozo will just ship a "disable the sandbox" override in their module and increase atrack surface for everyone. Or two pam modules will have conflicting overrides.

I would also be fine with FD passing or connection to a broker daemon.

jrybar-rh:
> > It is obvious that issues of this type get more and more numerous every week. I accept that, because removing SUID bit from polkit-agent-helper is a milestone that looks like a hurdle for some. Partially it is. Objectively, security should be paramount in polkit as an authorization suite. Some might call that an inconvenience. Even SELinux/Apparmor might look like a nuisance, but your distro certainly got along with it. It requires a bit more energy to make it work seamlessly, but it is worth the security it should provide.
> > I don't doubt there might be a process that can straightforward the experience this change brings, maybe even a new system-wide concept, and it only needs time to find this process.
> > I thought that it might make sense to leave this to the distro package maintainers, i.e. the drop-in overrides should be parts of the packages that need the override, distro-specific. Making changes in the unit file here on polkit upstream doesn't make sense. The unit file would turn into a list of exceptions for all the projects that use PAM modules and interact with polkit.
> > @bluca, does it make sense?
> 
> Yes this makes perfect sense, it's up to the pam modules to ship the drop in overrides they need. Making everything worse for everyone by default is definitely not the best solution. Shipping a drop in override is trivial and any pam package can do it easily, and they are in the best position to know what access their module actually needs, and doesn't need to wait for a new polkit release every time they change something.

I was speaking of overrides distributed by the distros instead of the PAM module upstreams. Not every distro uses systemd, while PAM modules are init-agnostic. To me, it makes sense that the creators of the distro compose know best what configuration should be used and whether or not a package should distribute an override. In short, PAM module creators do not know who uses them, while distro maintainers know what they use. I plan to mention this in the manpage. Please tell me if it's total nonesense.

jrybar-rh:
Closing after a discussion that it shouldn't be fixed at the polkit upstream level.

bluca:
> To me, it makes sense that the creators of the distro compose know best what configuration should be used and whether or not a package should distribute an override. In short, PAM module creators do not know who uses them, while distro maintainers know what they use. I plan to mention this in the manpage. Please tell me if it's total nonesense.

No, on the contrary, these need to be shipped and maintained by the PAM module creators, as they know what access is required, and when they change something. It's definitely not something for packagers to do, as it has absolutely nothing to do with packaging, it's completely unrelated.

jrybar-rh:
> > To me, it makes sense that the creators of the distro compose know best what configuration should be used and whether or not a package should distribute an override. In short, PAM module creators do not know who uses them, while distro maintainers know what they use. I plan to mention this in the manpage. Please tell me if it's total nonesense.
> 
> No, on the contrary, these need to be shipped and maintained by the PAM module creators, as they know what access is required, and when they change something. It's definitely not something for packagers to do, as it has absolutely nothing to do with packaging, it's completely unrelated.

I will incorporate this into the wording of the manpage extension.

PR #672, opened 2026-06-12
pkexec: allow callers to preserve some safe needed vars

### LINKED ISSUE #671: [RFE] Consider allowing pkexec to let callers preserve some env vars
I'd like to suggest adding a new annotation to allow keeping some additional variables from the callers environment. 

Currently pkexec sanitizes the vars and uses a predefined allowlist passed to processes (for good reason) and there is no current way to preserve other variables that some programs like `mock` might need. Really this is to aid in moving mock away from consolehelper. we've tried to work without this on Amazon Linux but the only viable route is sudo and thats too broad of an authorisation to give. 

I tried `pkexec env VAR mock` for example but it still is a root level command and targets `env` as the authorised program. So theres a loss of pre-program auth scope with mock here. Plus it kinda defeats the sanitization already done as theres no argv validation

I think it would be helpful to combine the per action invocation in policies with sudo features like keep_env. I Will attach an accompanying PR for review

Thanks for considering. 

**Additional context**
Mock ask: https://github.com/rpm-software-management/mock/issues/684
Fedora bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=502765 for their moving away from usemode to polkit

rajibade:
The current process doesn't allow passing of some useful vars for certain processes like `mock` which relies on sudo and sudo-like tools to keep some. Lets allowlist some while still being secure. Mirroring how allow_gui is done here and using glibc AT_SECURE for denylisting. I'm not sure however if policy files are checked to be root owned which is what this relies on so they can't be changed easily. 

Fixes: #671

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -88,6 +88,7 @@ check_functions = [
   'clearenv',
   'fdatasync',
   'setnetgrent',
+  'getauxval',
 ]
 
 foreach func: check_functions
--- a/src/programs/pkexec.c
+++ b/src/programs/pkexec.c
@@ -28,6 +28,7 @@
 #include <grp.h>
 #include <pwd.h>
 #include <errno.h>
+#include <sys/auxv.h>
 
 #ifdef __linux__
 #include <sys/prctl.h>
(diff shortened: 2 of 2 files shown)


rajibade:
Hi, can I poke someone to take a review at this?

jrybar-rh:
It's in the backlog. We're reviewing as much as we can. Please be patient.

PR #673, opened 2026-06-16
fix: reject UID/GID values exceeding G_MAXINT32 to prevent truncation

xuammu:
● fix: reject UID/GID values exceeding G_MAXUINT32 to prevent integer truncation

  ## Summary

  polkit_identity_from_string() parses unix-user: and unix-group: identity strings by converting the numeric portion from guint64 to gint with no bounds check. Values where val % 2^32 == 0 wrap to uid/gid 0
  (root): unix-user:4294967296, 8589934592, 12884901888 all produce uid=0.

  ## Fix

  Add errno=0 before g_ascii_strtoull() and check errno == ERANGE || val > G_MAXUINT32 before the (gint) cast in both the unix-user: and unix-group: branches. Out-of-range values set a GError instead of
  producing a truncated identity. Also add #include <errno.h> which was missing.

  G_MAXUINT32 (not G_MAXINT) is the correct bound because uid_t is a 32-bit unsigned type. Values in (G_MAXINT, G_MAXUINT32] are valid high UIDs — polkit's own test data uses UID 4000000000.

  ## Tests

  Added 10 edge-value tests in polkitidentitytest.c: 6 rejection tests (wrap-to-root and G_MAXUINT32 for both uid/gid) and 4 acceptance tests (G_MAXINT and G_MAXINT+1 for both uid/gid).

  ## Affected code

  src/polkit/polkitidentity.c, test/polkit/polkitidentitytest.c

### DIFF
--- a/src/polkit/polkitidentity.c
+++ b/src/polkit/polkitidentity.c
@@ -19,6 +19,7 @@
  * Author: David Zeuthen <davidz@redhat.com>
  */
 
+#include <errno.h>
 #include <string.h>
 
 #include "polkitidentity.h"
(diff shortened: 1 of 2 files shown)


jrybar-rh:
@xuammu Can you please update the PR name, summary and other metadata to actually contain MAXUINT? Thank you. And also can you please add the edge values in the unit tests? 
Many thanks.

jrybar-rh:
Hello @xuammu,
Thanks for the commit. 
The tests revealed that the new PolkitIdentity test is failing:
```
ok 12 /PolkitIdentity/uid_overflow_wrap_to_root
ok 13 /PolkitIdentity/uid_overflow_wrap_to_root_2
not ok /PolkitIdentity/uid_overflow_maxuint32 - FATAL-CRITICAL: polkit_unix_user_new: assertion 'uid
 != -1' failed
Bail out!
```

Can you please look into this?
Thank you.

PR #674, opened 2026-06-19
polkitidentity: reject UID/GID values exceeding G_MAXINT to prevent integer truncation

rawrmonster17:
## Summary

`polkit_identity_from_string()` parses `unix-user:` and `unix-group:` identity strings by converting the numeric portion from `guint64` to `gint` with no bounds check. This causes silent integer truncation for large values.

## The bug

Values where `val % 2³² == 0` wrap to uid/gid 0 (root):

```
unix-user:4294967296   → uid=0  (root)
unix-user:8589934592   → uid=0  (root)
unix-user:12884901888  → uid=0  (root)
```

Values in `(G_MAXINT, G_MAXUINT32]` wrap to negative `gint` values. `polkit_unix_user_new()` only guards against `-1`, so all other negative values are accepted as valid (non-root) but incorrect UIDs:

```
unix-user:4294967295  → (gint)-1  → NULL   (blocked by g_return_val_if_fail)
unix-user:4294967294  → (gint)-2  → accepted as uid -2
unix-user:2147483648  → (gint)-2147483648 → accepted
```

## Why G_MAXINT is the correct bound

The `uid` and `gid` fields in `PolkitUnixUser`/`PolkitUnixGroup` are declared as `gint`. The GObject property spec registers them with range `[G_MININT, G_MAXINT]`. Any numeric string value above `G_MAXINT` (2,147,483,647) must be rejected — not `G_MAXUINT32`, which still leaves the entire `(G_MAXINT, G_MAXUINT32]` range producing negative non-(-1) gints that pass through.

PR #673 used `G_MAXUINT32` as the bound, which is insufficient for this reason.

## The fix

Add a `val > (guint64) G_MAXINT` check before the cast in both the `unix-user:` and `unix-group:` branches. Out-of-range values now set a `GError` instead of producing a truncated identity.

Also adds six regression tests to `polkitidentitytest.c` covering:
- The three wrap-to-root cases (`4294967296`, `8589934592`)
- The `G_MAXINT+1` boundary (`2147483648`)
- The `G_MAXUINT32` boundary (`4294967295`)
- GID equivalents

All tests pass. `unix-user:0` (root) and all valid UIDs in `0..G_MAXINT` are unaffected.

## Note on related path

The `polkit_identity_new_for_gvariant()` path deserializes uid/gid as `guint32` → `gint` via `g_variant_get_uint32()`. That path is used over D-Bus (not from string parsing) and warrants a separate audit — out of scope here.

### DIFF
--- a/src/polkit/polkitidentity.c
+++ b/src/polkit/polkitidentity.c
@@ -20,6 +20,7 @@
  */
 
 #include <string.h>
+#include <errno.h>
 
 #include "polkitidentity.h"
 #include "polkitunixuser.h"
@@ -156,22 +157,46 @@ polkit_identity_from_string  (const gchar   *str,
 
   if (g_str_has_prefix (str, "unix-user:"))
     {
+      errno = 0;
       val = g_ascii_strtoull (str + sizeof "unix-user:" - 1,
                               &endptr,
                               10);
       if (*endptr == '\0')
-        identity = polkit_unix_user_new ((gint) val);
+        {
+          if (errno == ERANGE || val > (guint64) G_MAXUINT32)
+            {
+              g_set_error (error,
+                           POLKIT_ERROR,
+                           POLKIT_ERROR_FAILED,
+                           "UID value '%s' is out of range",
+                           str + sizeof "unix-user:" - 1);
+            }
+          else
+            identity = polkit_unix_user_new ((gint) val);
+        }
       else
         identity = polkit_unix_user_new_for_name (str + sizeof "unix-user:" - 1,
                                                  error);
     }
   else if (g_str_has_prefix (str, "unix-group:"))
     {
+      errno = 0;
       val = g_ascii_strtoull (str + sizeof "unix-group:" - 1,
                               &endptr,
                               10);
       if (*endptr == '\0')
-        identity = polkit_unix_group_new ((gint) val);
+        {
+          if (errno == ERANGE || val > (guint64) G_MAXUINT32)
+            {
+              g_set_error (error,
+                           POLKIT_ERROR,
+                           POLKIT_ERROR_FAILED,
+                           "GID value '%s' is out of range",
+                           str + sizeof "unix-group:" - 1);
+            }
+          else
+            identity = polkit_unix_group_new ((gint) val);
+        }
       else
         identity = polkit_unix_group_new_for_name (str + sizeof "unix-group:" - 1,
                                                   error);
--- a/test/polkit/polkitidentitytest.c
+++ b/test/polkit/polkitidentitytest.c
@@ -134,6 +134,37 @@ test_comparison (const void *_data)
 }
 
 
+static void
+test_overflow_reject (const void *_subject)
+{
+  const gchar *subject = (const gchar *) _subject;
+
+  PolkitIdentity *identity;
+  GError *error = NULL;
+
+  /* Out-of-range numeric uid/gid must be rejected with an error, not silently truncated */
+  identity = polkit_identity_from_string (subject, &error);
+  g_assert (identity == NULL);
+  g_assert_error (error, POLKIT_ERROR, POLKIT_ERROR_FAILED);
+  g_clear_error (&error);
+}
+
+
+static void
+test_overflow_accept (const void *_subject)
+{
+  const gchar *subject = (const gchar *) _subject;
+  PolkitIdentity *identity;
+  GError *error = NULL;
+
+  /* Values at or below G_MAXINT must be accepted */
+  identity = polkit_identity_from_string (subject, &error);
+  g_assert (error == NULL);
+  g_assert (identity != NULL);
+  g_object_unref (identity);
+}
+
+
 /* Test helpers */
 
 struct ComparisonTestData comparison_test_data [] = {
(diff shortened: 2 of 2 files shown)


jrybar-rh:
## Code Review: PR #674 — polkitidentity: reject UID/GID values exceeding G_MAXINT

**Files reviewed:** `src/polkit/polkitidentity.c`, `test/polkit/polkitidentitytest.c`
**Context also read:** `src/polkit/polkitunixuser.c`, `src/polkit/polkitunixgroup.c`

The PR correctly identifies and fixes a real integer truncation vulnerability in `polkit_identity_from_string()`. The core fix — guarding `val > (guint64) G_MAXINT` before the `(gint)` cast — is sound, and the choice of `G_MAXINT` (rather than `G_MAXUINT32` as in #673) is the right bound given that `PolkitUnixUser.uid` and `PolkitUnixGroup.gid` are declared as `gint`. The fix prevents truncation-to-root attacks for values like `4294967296`.

---

### Medium

**`src/polkit/polkitidentity.c` — no `errno`/overflow check on `g_ascii_strtoull` itself**

`g_ascii_strtoull` (like POSIX `strtoull`) sets `errno = ERANGE` and returns `G_MAXUINT64` when the input overflows `guint64` (e.g., `"unix-user:99999999999999999999"`). The PR does not clear or check `errno` before calling `g_ascii_strtoull`.

Because `G_MAXUINT64 > (guint64)G_MAXINT`, the overflow case *is* caught by the new guard and rejected — so there is no security regression. However, the error message then misleadingly reports the clamped value (`18446744073709551615`) rather than the actual string the caller supplied. This can confuse log analysis and debugging.

Suggested fix:
```c
errno = 0;
val = g_ascii_strtoull (str + sizeof "unix-user:" - 1, &endptr, 10);
if (*endptr == '\0')
  {
    if (errno == ERANGE || val > (guint64) G_MAXINT)
      {
        g_set_error (error,
                     POLKIT_ERROR,
                     POLKIT_ERROR_FAILED,
                     "UID value '%s' is out of range",
                     str + sizeof "unix-user:" - 1);
      }
    else
      identity = polkit_unix_user_new ((gint) val);
  }
```
The same applies to the `unix-group:` branch. Adding `#include <errno.h>` is also required since it is not currently included in `polkitidentity.c`.

---

### Low

**`test/polkit/polkitidentitytest.c` — asymmetric GID test coverage**

The UID overflow tests cover four cases: `4294967296` (wrap-to-root), `8589934592` (second wrap-to-root), `2147483648` (G_MAXINT+1), `4294967295` (G_MAXUINT32). The GID tests only cover two: `4294967296` and `2147483648`. The cases `8589934592` (second wrap-to-root for GID) and `4294967295` (G_MAXUINT32 for GID) are not tested.

Suggested addition alongside the existing GID tests:
```c
g_test_add_data_func ("/PolkitIdentity/gid_overflow_wrap_to_root_2",
                      "unix-group:8589934592", test_overflow_reject);
g_test_add_data_func ("/PolkitIdentity/gid_overflow_maxuint32",
                      "unix-group:4294967295", test_overflow_reject);
```

---

### Low

**`test/polkit/polkitidentitytest.c` — no positive boundary test for `G_MAXINT`**

The tests verify that `G_MAXINT+1` is rejected but do not verify that `G_MAXINT` itself (`2147483647`) is accepted. If the bound were accidentally written as `>=` instead of `>`, the positive boundary would be silently broken. A test for the exact boundary strengthens confidence:
```c
g_test_add_data_func ("/PolkitIdentity/uid_maxint_valid",
                      "unix-user:2147483647", test_overflow_reject /* use a test_accepts helper */);
```
(Requires a small `test_accepts` helper mirroring `test_overflow_reject` but asserting `identity != NULL && error == NULL`.)

---

### Nitpick

**`test/polkit/polkitidentitytest.c` — `test_overflow_reject` does not assert the error domain/code**

The test asserts `error != NULL` but does not verify that the error is `POLKIT_ERROR`/`POLKIT_ERROR_FAILED`. A future refactor that accidentally sets a different GError would still pass this test. Consider adding:
```c
g_assert_error (error, POLKIT_ERROR, POLKIT_ERROR_FAILED);
```
in place of `g_assert (error != NULL)`.

---

### Out-of-scope note (acknowledged in PR description)

The `polkit_identity_new_for_gvariant()` path (lines 325–328 and 341–344 of `polkitidentity.c`) performs a `guint32 → gint` implicit conversion when deserializing UID/GID from D-Bus variants. Values in `(G_MAXINT, G_MAXUINT32]` will wrap to negative `gint` values that are accepted by `polkit_unix_user_new()` (since only `-1` is guarded). This pre-existing truncation is correctly called out by the PR as warranting a separate audit.

---

### Summary

The fix is correct and closes the reported vulnerability. The two Medium/Low concerns are about defensive completeness (errno check, test symmetry) rather than correctness of the security fix itself. Recommend addressing the `errno` check and test gaps before merging.

---

rawrmonster17:
Thanks for the thorough review. All points addressed in the updated commit:

**Medium — errno/ERANGE check:**
Added `errno = 0` before each `g_ascii_strtoull` call and updated the bounds check to `if (errno == ERANGE || val > (guint64) G_MAXINT)`. Also added `#include <errno.h>` which was missing. Error message now reports the original input string rather than the clamped `G_MAXUINT64` value, so logs are accurate on overflow.

**Low — asymmetric GID test coverage:**
Added the two missing GID cases: `gid_overflow_wrap_to_root_2` (`unix-group:8589934592`) and `gid_overflow_maxuint32` (`unix-group:4294967295`). GID and UID test sets are now symmetric at 4 cases each.

**Low — no G_MAXINT acceptance test:**
Added a `test_overflow_accept` helper (asserts `identity != NULL && error == NULL`, unrefs) and two boundary tests: `uid_maxint_valid` (`unix-user:2147483647`) and `gid_maxint_valid` (`unix-group:2147483647`). These catch an accidental `>=` vs `>` regression.

**Nitpick — error domain/code not asserted:**
`g_assert_error(error, POLKIT_ERROR, POLKIT_ERROR_FAILED)` replaces the bare `g_assert(error != NULL)` in `test_overflow_reject`. `POLKIT_ERROR` is available via the existing `<polkit/polkit.h>` umbrella include.

rawrmonster17:
CI failure found and fixed — updated commit pushed.

**Root cause of the failure:**

The build compiled cleanly but `meson test` failed in `test-polkitbackendjsauthority`:

```
ERROR: test-polkitbackendjsauthority.c:447: assertion failed (error == NULL):
UID value '2147483648' is out of range (polkit-error-quark, 0)
```

The backend test feeds `unix-user:2147483648` (UID `highuid1` in `test/data/etc/passwd`) and expects `NOT_AUTHORIZED` — meaning the identity should parse successfully but be denied. Our bound of `G_MAXINT` was wrong: it rejected `2147483648` at parse time instead of letting it through.

**The correct bound is `G_MAXUINT32`, not `G_MAXINT`.**

`uid_t` is a 32-bit unsigned type. Values in `(G_MAXINT, G_MAXUINT32]` are valid 32-bit UIDs in real use — polkit's own test data includes `highuid2` at UID `4000000000`. Only values above `G_MAXUINT32` (4294967296+) are true 64-bit overflows that wrap to root and must be rejected.

The original PR description's claim that "G_MAXINT is the correct bound" was incorrect. G_MAXUINT32 is the right limit — it closes the wrap-to-root attack vector while preserving compatibility with high UIDs.

**What changed in the updated commit:**
- Bound changed from `G_MAXINT` to `G_MAXUINT32` in both the `unix-user:` and `unix-group:` branches
- `uid_overflow_maxint_plus1` and `gid_overflow_maxint_plus1` test cases renamed to `uid_highuid_maxint_plus1` / `gid_highuid_maxint_plus1` and changed to `test_overflow_accept` — `2147483648` is a valid UID
- All other test cases unchanged: `4294967296` (wrap-to-root) still rejected, `4294967295` still rejected (casts to -1, blocked by `polkit_unix_user_new`'s existing guard), `8589934592` still rejected

PR #675, opened 2026-06-19
polkitbackend: fix NULL deref crash (SIGSEGV) in push_subject() GError contract violation

### LINKED ISSUE #670: polkitd crash (SIGSEGV) due to GError contract violation in push_subject()
I played with results of https://svashisht.fedorapeople.org/openscanhub/mass-scans/ai-analysis-f45-28-Apr-2026/polkit/ and it seems this is a real issue, (although you need to have specific rule file)
 

Summary
-------
polkitd crashes with SIGSEGV when push_subject() fails to find the JS
"Subject" global constructor. The function returns FALSE without setting
the GError output parameter, but both callers unconditionally dereference
error->message in their error-logging path. This is a NULL pointer
dereference that crashes the polkitd authorization daemon.

A secondary, lower-severity variant of the same pattern exists in
polkit_authority_get() (client-side, deprecated API).

Affected Version
----------------
polkit 127 (current main branch as of 2026-05-04)
Also confirmed on polkit-126-2.fc42 (Fedora 42)
The code has been present since push_subject() was introduced.

Reproduction
------------
Tested on Fedora 42, polkit-126-2.fc42.x86_64.

    # 1. Create a rules file that removes the Subject constructor
    echo 'delete Subject;' > /etc/polkit-1/rules.d/00-delete-subject.rules

    # 2. Restart polkitd
    systemctl restart polkit

    # 3. Trigger an authorization check
    useradd -m testuser
    su - testuser -c 'pkexec echo hello'

Result: polkitd crashes with SIGSEGV.

    $ journalctl -u polkit --since '1 minute ago'
    polkit.service: Main process exited, code=dumped, status=11/SEGV
    polkit.service: Failed with result 'core-dump'.

Backtrace from coredumpctl:

    #0  polkit_backend_common_js_authority_check_authorization_sync
        at polkitbackendduktapeauthority.c:1020
    #1  polkit_backend_interactive_authority_check_authorization_sync
        at polkitbackendinteractiveauthority.c:1399
    #2  check_authorization_sync
        at polkitbackendinteractiveauthority.c:1172
    #3  polkit_backend_interactive_authority_check_authorization
        at polkitbackendinteractiveauthority.c:989
    #4  polkit_backend_authority_check_authorization
        at polkitbackendauthority.c:237
    #5  server_handle_check_authorization
        at polkitbackendauthority.c:846
    #6  server_handle_method_call
        at polkitbackendauthority.c:1328

Cleanup: rm /etc/polkit-1/rules.d/00-delete-subject.rules && systemctl restart polkit

Root Cause
----------
push_subject() in polkitbackendduktapeauthority.c has an early-return
path that violates the GError contract:

    /* line 394-396 */
    if (!duk_get_global_string (cx, "Subject")) {
      return FALSE;    /* <-- returns failure without setting *error */
    }

Both callers unconditionally dereference error->message when
push_subject() returns FALSE:

    /* line 940-947 (get_admin_auth_identities) */
    if (!push_subject (cx, subject, ..., &error))
      {
        polkit_backend_authority_log (POLKIT_BACKEND_AUTHORITY (authority),
                                      LOG_LEVEL_ERROR,
                                      "Error converting subject to JS object: %s",
                                      error->message);    /* NULL deref */
        g_clear_error (&error);
        goto out;
      }

    /* line 1022-1029 (check_authorization_sync) -- identical pattern */
    if (!push_subject (cx, subject, ..., &error))
      {
        polkit_backend_authority_log (POLKIT_BACKEND_AUTHORITY (authority),
                                      LOG_LEVEL_ERROR,
                                      "Error converting subject to JS object: %s",
                                      error->message);    /* NULL deref */
        g_clear_error (&error);
        goto out;
      }

GError *error is initialized to NULL at the top of each caller function
(line 917 and 1002 respectively). When push_subject() takes the line 395
early-return path, error stays NULL, and error->message is a NULL pointer
dereference.

Impact
------
- polkitd is a privileged root-running system daemon
- The crash is reachable via the D-Bus CheckAuthorization and
  EnumerateTemporaryAuthorizations paths
- During the crash, all authorization checks system-wide fail
- systemd auto-restarts polkitd, but repeated triggers cause a crash
  loop that effectively denies all privilege escalation
- The trigger condition (JS Subject global missing) can occur due to
  Duktape engine initialization failure, JS heap exhaustion, or
  malformed rules files

Suggested Fix
-------------
Fix push_subject() to set *error before returning FALSE (preferred,
follows the GLib GError contract):

    --- a/src/polkitbackend/polkitbackendduktapeauthority.c
    +++ b/src/polkitbackend/polkitbackendduktapeauthority.c
    @@ -393,7 +393,10 @@ push_subject (duk_context               *cx,

       if (!duk_get_global_string (cx, "Subject")) {
    -    return FALSE;
    +    g_set_error (error,
    +                 POLKIT_ERROR,
    +                 POLKIT_ERROR_FAILED,
    +                 "JS global 'Subject' not found in Duktape context");
    +    return FALSE;
       }

As a defense-in-depth measure, also guard the dereference sites:

    --- a/src/polkitbackend/polkitbackendduktapeauthority.c
    +++ b/src/polkitbackend/polkitbackendduktapeauthority.c
    @@ -942,7 +942,7 @@ polkit_backend_common_js_authority_get_admin_auth_identities (...)
           polkit_backend_authority_log (POLKIT_BACKEND_AUTHORITY (authority),
                                         LOG_LEVEL_ERROR,
                                         "Error converting subject to JS object: %s",
    -                                    error->message);
    +                                    error != NULL ? error->message : "unknown error");

(Same change at line 1027 in check_authorization_sync.)

Related: push_action_and_details() has the same bare return FALSE
-----------------------------------------------------------------
push_action_and_details() at line 571-572 also returns FALSE without
setting a GError when duk_get_global_string(cx, "Action") fails.
Its callers (lines 932 and 1014) log a static string without
dereferencing error->message, so they are not currently vulnerable.
However, this is still a GError contract violation and should be
fixed for consistency:

    /* line 571-572 */
    if (!duk_get_global_string (cx, "Action")) {
      return FALSE;    /* should also set *error -- but this function
                          does not take a GError** parameter at all */
    }

Since push_action_and_details() does not accept a GError** parameter,
it cannot set an error. The callers correctly handle this by not
attempting to read error->message. No crash here, but consider whether
this function should accept GError** for future callers.

Lower-severity: polkit_authority_get() in polkitauthority.c
-----------------------------------------------------------
polkit_authority_get() (deprecated) also dereferences error->message
without a NULL guard:

    /* polkitauthority.c line 413-420 */
    error = NULL;
    ret = polkit_authority_get_sync (NULL, &error);
    if (ret == NULL)
      {
        g_warning ("polkit_authority_get: Error getting authority: %s",
                   error->message);    /* NULL deref if error unset */
        g_error_free (error);
      }

This is reachable only if get_uninitialized_authority() returns NULL
without setting *error (e.g., g_object_new failure). In modern GLib
(>= 2.68), g_object_new aborts on OOM, so this is nearly impossible.
The impact is client-side only (not polkitd). Fix:

    g_warning ("polkit_authority_get: Error getting authority: %s",
               error != NULL ? error->message : "unknown error");
    g_clear_error (&error);

Discovery
---------
Found by GCC -fanalyzer (GCC 16.0.1) via OpenScanHub mass-scan of
Fedora 45, confirmed by manual source review and runtime reproduction
with SIGSEGV + coredump on Fedora 42.

This report was prepared with AI assistance and should be verified by
a human reviewer.

rawrmonster17:
Fixes #670

## The crash

`push_subject()` returns `FALSE` without setting `*error` when `duk_get_global_string(cx, "Subject")` fails. Both callers then unconditionally dereference `error->message` on the failure path — NULL pointer dereference, SIGSEGV, polkitd dies.

`GError *error` is initialized to `NULL` at the top of each caller. When `push_subject()` takes the early-return path at line 394, `error` stays `NULL` and `error->message` crashes the privileged daemon.

Reproduced on Fedora 42 (`polkit-126-2.fc42`) by placing `delete Subject;` in a rules file and triggering any authorization check. The trigger condition can also occur due to Duktape engine initialization failure or JS heap exhaustion.

## What this patch does

**Primary fix — `polkitbackendduktapeauthority.c`:**

`push_subject()` now calls `g_set_error()` before returning `FALSE`, satisfying the GError contract. The callers receive a valid error object and log it correctly.

**Defense-in-depth — both `push_subject()` call sites:**

Both `error->message` dereferences are guarded with `error != NULL ? error->message : "(no error message)"`. This makes the log path safe even if another code path regresses in the future.

**Secondary fix — `polkitauthority.c`:**

`polkit_authority_get()` (deprecated client-side API) had the same unguarded `error->message` dereference. Applied the same NULL guard and replaced `g_error_free(error)` with `g_clear_error(&error)`, which is NULL-safe.

## Diff summary

```
src/polkitbackend/polkitbackendduktapeauthority.c | +4 lines (g_set_error in push_subject)
                                                   | +2 lines (NULL guards on both call sites)
src/polkit/polkitauthority.c                       | +1 line  (NULL guard + g_clear_error)
```

2 files, 8 insertions, 4 deletions, 1 commit.

### DIFF
--- a/src/polkit/polkitauthority.c
+++ b/src/polkit/polkitauthority.c
@@ -416,8 +416,8 @@ polkit_authority_get (void)
   if (ret == NULL)
     {
       g_warning ("polkit_authority_get: Error getting authority: %s",
-                 error->message);
-      g_error_free (error);
+                 error != NULL ? error->message : "(no error message)");
+      g_clear_error (&error);
     }
 
   return ret;
--- a/src/polkitbackend/polkitbackendduktapeauthority.c
+++ b/src/polkitbackend/polkitbackendduktapeauthority.c
@@ -392,6 +392,11 @@ push_subject (duk_context               *cx,
   char *system_unit = NULL;
 
   if (!duk_get_global_string (cx, "Subject")) {
+    duk_pop (cx); /* duk_get_global_string pushes undefined on failure; discard it */
+    g_set_error (error,
+                 POLKIT_ERROR,
+                 POLKIT_ERROR_FAILED,
+                 "JS global 'Subject' constructor not found in Duktape context");
     return FALSE;
   }
 
@@ -569,6 +574,7 @@ push_action_and_details (duk_context               *cx,
   guint n;
 
   if (!duk_get_global_string (cx, "Action")) {
+    duk_pop (cx); /* duk_get_global_string pushes undefined on failure; discard it */
     return FALSE;
   }
 
@@ -942,7 +948,7 @@ polkit_backend_common_js_authority_get_admin_auth_identities (PolkitBackendInter
       polkit_backend_authority_log (POLKIT_BACKEND_AUTHORITY (authority),
                                     LOG_LEVEL_ERROR,
                                     "Error converting subject to JS object: %s",
-                                    error->message);
+                                    error != NULL ? error->message : "(no error message)");
       g_clear_error (&error);
       goto out;
     }
@@ -1024,7 +1030,7 @@ polkit_backend_common_js_authority_check_authorization_sync (PolkitBackendIntera
       polkit_backend_authority_log (POLKIT_BACKEND_AUTHORITY (authority),
                                     LOG_LEVEL_ERROR,
                                     "Error converting subject to JS object: %s",
-                                    error->message);
+                                    error != NULL ? error->message : "(no error message)");
       g_clear_error (&error);
       goto out;
     }


jrybar-rh:
## Code Review — PR #675: fix NULL deref crash (SIGSEGV) in push_subject() GError contract violation

Reviewed files:
- `src/polkitbackend/polkitbackendduktapeauthority.c`
- `src/polkit/polkitauthority.c`

---

### Summary

The PR correctly identifies and fixes two distinct NULL-dereference bugs: (1) `push_subject()` violating the GError contract by returning `FALSE` without populating `*error`, and (2) `polkit_authority_get()` unconditionally dereferencing `error->message` on the failure path of the deprecated client API. Both fixes are directionally correct and address real crashes. There are however some issues worth noting below.

---

## Issues by Severity

### Medium

**1. `polkitauthority.c:416–421` — NULL-error path for `polkit_authority_get()` is theoretically impossible at this call site**

`get_uninitialized_authority()` (line 428–447) never sets an error — it either returns the cached singleton or allocates a new object via `g_object_new()` (which can only fail with `g_error()`/abort, not via GError). The only place `error` can be set is `g_initable_init()`, which always sets `*error` on failure according to GLib's GInitable contract (backed by `g_return_val_if_fail (error == NULL || *error == NULL, FALSE)` in GLib internals and documented semantics).

So the `error != NULL ?` guard in `polkitauthority.c` is actually dead code at this specific call site — but it is harmless and provides future-proofing. Still worth a comment explaining why the guard is needed, so a future reader understands this isn't cargo-cult code.

**2. `polkitbackendduktapeauthority.c:394–396` — Duktape stack leak on early return**

Before the `g_set_error()` + `return FALSE` path added by this PR, `duk_get_global_string(cx, "Subject")` was already called. According to Duktape documentation, `duk_get_global_string()` pushes a value onto the stack regardless of whether it returns true or false (it pushes `undefined` on failure). The existing code at line 398 calls `duk_new(cx, 0)` only on the success path, but the false-return path now leaks the `undefined` value on the Duktape stack. This could cause stack imbalance across subsequent JS evaluations.

The fix should include `duk_pop(cx)` before returning:
```c
if (!duk_get_global_string (cx, "Subject")) {
    duk_pop (cx); /* discard the undefined value pushed by duk_get_global_string */
    g_set_error (error,
                 POLKIT_ERROR,
                 POLKIT_ERROR_FAILED,
                 "JS global 'Subject' constructor not found in Duktape context");
    return FALSE;
}
```
Note: this Duktape stack imbalance was a pre-existing bug exposed by the current understanding of the code flow — it's worth fixing as part of this patch since you're already touching this exact line.

**3. `push_action_and_details()` at line 571 has the same missing `duk_pop()` pattern**

`push_action_and_details()` has an identical `duk_get_global_string(cx, "Action")` + `return FALSE` without `duk_pop()`. While out of scope for this PR's stated fix, the same stack-leak concern applies. Both functions should be fixed together for consistency.

---

### Low

**4. `polkitbackendduktapeauthority.c` — `push_subject()` NULL guard at callers is now redundant (but not harmful)**

After the primary fix (adding `g_set_error()` in `push_subject()`), the `error != NULL ?` guards added at lines 945 and 1027 are defensive against a regression but not necessary for correctness. The GError contract for `push_subject()` is now satisfied. The guards are fine to keep as defense-in-depth, but a comment would help, e.g.:
```c
/* Defensive NULL guard: push_subject() should always set *error on failure,
 * but guard here to avoid a crash if that contract is ever broken again. */
```

**5. `polkitauthority.c` — `g_clear_error()` is correct, but reason for the change deserves a comment**

Replacing `g_error_free(error)` with `g_clear_error(&error)` is the right call — `g_clear_error()` is NULL-safe and also zeroes the pointer (preventing use-after-free). This is a good cleanup. A brief commit-message note would suffice; no in-code comment needed.

---

### Nitpick

**6. Error message string in `g_set_error()` could be more operator-friendly**

```c
"JS global 'Subject' constructor not found in Duktape context"
```

This message is good for developers but may be cryptic in syslog for operators. Something like:

```c
"polkit rules engine: JS 'Subject' constructor not found; the rules engine may be corrupted"
```

This is purely stylistic and consistent with how other messages in this file are written (e.g., `"Error deleting old rules, not loading new ones"`).

---

## What the PR Does Well

- The root cause analysis in the PR description is accurate and well-explained.
- Using `POLKIT_ERROR` / `POLKIT_ERROR_FAILED` is correct for the domain.
- `g_clear_error()` replacing `g_error_free()` in `polkitauthority.c` is the right GLib idiom.
- The fix is minimal and surgical — no unnecessary refactoring.

---

## Verdict

The fix correctly addresses the SIGSEGV. The Duktape stack leak (issue #2) is the most substantive concern and should be resolved before merge. Issues #1 and #3–6 are lower priority but worth addressing for correctness and maintainability in a security-critical daemon.

---

rawrmonster17:
Thanks for the review. All points addressed in the updated commit:

**Medium — Duktape stack leak on early return (push_subject and push_action_and_details):**
Both `duk_get_global_string` failure paths now call `duk_pop(cx)` before returning to discard the `undefined` value that Duktape pushes even on failure. `push_subject` gets `duk_pop` + `g_set_error` + `return FALSE`. `push_action_and_details` gets just `duk_pop` + `return FALSE` since it takes no `GError**`.

**Medium — polkitauthority.c NULL guard is theoretically dead code:**
Agreed the `error` path is currently unreachable at that call site given GLib's GInitable contract. The guard is kept as future-proofing — if the call chain ever changes it won't silently crash. The `g_error_free` → `g_clear_error` change stands regardless since `g_clear_error` is the correct idiom (NULL-safe, zeroes the pointer).

**Low — defensive NULL guards at call sites:**
Kept as-is. They're defense-in-depth against the same contract violation being re-introduced, consistent with how #670 was originally triggered.

PR #676, opened 2026-06-19
security: fix three hardening issues found by static analysis

rawrmonster17:
Three independent fixes found by static analysis and manual source audit. One commit, three files.

---

## Fix 1 — `polkit_unix_process_set_pid()` opens pidfd for the wrong pid (`polkitunixprocess.c`)

**Bug:** `polkit_unix_process_set_pid()` called `syscall(SYS_pidfd_open, process->pid, 0)` — passing the stale *stored* value rather than the `pid` parameter. `process->pid = pid` only happens on the non-pidfd fallback path at the bottom of the function (line 870).

In practice during normal object construction, `process->pid` is 0 at the point `set_pid` is called (GObject zero-initialises the struct; `polkit_unix_process_init()` does not set `pid`). So `pidfd_open(0)` fails with `ESRCH`, the fallback path is taken, and `constructed()` subsequently opens the correct pidfd — masking the bug. The latent risk is a post-construction caller that sets a new pid on an existing object: it would attempt to open a pidfd for the old/stale pid.

**Fix:** pass `pid` (the parameter) to `SYS_pidfd_open`. This matches the correct pattern already used in `polkit_unix_process_constructed()`.

---

## Fix 2 — `read_cookie()` 1-byte OOB read + `send_dbus_message()` unguarded `error->message` (`polkitagenthelperprivate.c`)

**Bug A — OOB read:** `read_cookie()` indexed `buf[strlen(buf) - 1]` with no zero-length guard. If the first byte delivered on stdin is `\0`, `strlen(buf)` returns 0 and `buf[-1]` is a one-byte read below the stack buffer. This is undefined behaviour and is caught by `-fsanitize=undefined`.

**Fix A:** store `strlen(buf)` in `len`, check `len == 0` before indexing.

**Bug B — NULL deref:** `send_dbus_message()` dereferenced `error->message` unconditionally on its failure path — the same GError contract violation pattern as the polkitd crash in #670/#675. `g_error_free()` was also used, which aborts on a NULL pointer.

**Fix B:** guard with `error != NULL ?` ternary; switch to `g_clear_error()` which is NULL-safe.

---

## Fix 3 — Action-id validation commented out; restore with new implementation (`polkitbackendactionpool.c`)

**Bug:** The validation call in `_start()` was commented out:
```c
//if (!polkit_action_validate_id (attr[1]))
//        goto error;
```
The referenced function `polkit_action_validate_id` no longer exists in the codebase. Malformed action IDs from `.policy` files (spaces, slashes, path traversal sequences, uppercase, etc.) were stored as-is and propagated into rule matching, D-Bus responses, and the JS `action.id` property.

**Fix:** Implement `_validate_action_id()` enforcing the documented reverse-DNS format: dot-separated components where each starts with a lowercase ASCII letter followed by lowercase letters, digits, or hyphens; no leading, trailing, or consecutive dots. Re-enable the check in `_start()` — invalid action IDs are now rejected at `.policy` load time with a `g_warning`.

---

## Summary

| # | File | Issue | Severity |
|---|------|-------|----------|
| 1 | `polkitunixprocess.c` | Wrong variable passed to `pidfd_open` | Low (latent) |
| 2a | `polkitagenthelperprivate.c` | 1-byte OOB read (`buf[-1]`) | Low |
| 2b | `polkitagenthelperprivate.c` | Unguarded `error->message` NULL deref | Low–Medium |
| 3 | `polkitbackendactionpool.c` | Missing action-id input validation | Low |

### DIFF
--- a/src/polkitagent/polkitagenthelperprivate.c
+++ b/src/polkitagent/polkitagenthelperprivate.c
@@ -68,7 +68,10 @@ read_cookie (int argc, char **argv)
             perror ("fgets");
           return NULL;
         }
-      if (buf[strlen (buf) - 1] != '\n')
+      /* Guard against a leading NUL byte causing strlen() to return 0,
+       * which would make buf[len-1] a one-byte read below the buffer. */
+      size_t len = strlen (buf);
+      if (len == 0 || buf[len - 1] != '\n')
         {
           /* Cookie too long - drain remaining input and reject */
           int c;
@@ -97,16 +100,18 @@ send_dbus_message (const char *cookie, const char *user, int pidfd, int uid)
   authority = polkit_authority_get_sync (NULL /* GCancellable* */, &error);
   if (authority == NULL)
     {
-      g_printerr ("Error getting authority: %s\n", error->message);
-      g_error_free (error);
+      g_printerr ("Error getting authority: %s\n",
+                  error != NULL ? error->message : "(no error message)");
+      g_clear_error (&error);
       goto out;
     }
 
   identity = polkit_unix_user_new_for_name (user, &error);
   if (identity == NULL)
     {
-      g_printerr ("Error constructing identity: %s\n", error->message);
-      g_error_free (error);
+      g_printerr ("Error constructing identity: %s\n",
+                  error != NULL ? error->message : "(no error message)");
+      g_clear_error (&error);
       goto out;
     }
 
@@ -128,8 +133,9 @@ send_dbus_message (const char *cookie, const char *user, int pidfd, int uid)
                                                               &error);
   if (!ret)
     {
-      g_printerr ("polkit-agent-helper-1: error response to PolicyKit daemon: %s\n", error->message);
-      g_error_free (error);
+      g_printerr ("polkit-agent-helper-1: error response to PolicyKit daemon: %s\n",
+                  error != NULL ? error->message : "(no error message)");
+      g_clear_error (&error);
       goto out;
     }
 
--- a/src/polkitbackend/polkitbackendactionpool.c
+++ b/src/polkitbackend/polkitbackendactionpool.c
@@ -737,6 +737,64 @@ pd_unref_data (ParserData *pd)
   pd->global_icon_name = NULL;
 }
 
+/* Validates a PolicyKit action identifier.
+ *
+ * Action IDs follow reverse-DNS notation: one or more dot-separated
+ * components where each component starts with an ASCII letter (upper or
+ * lower) and continues with ASCII letters, digits, or hyphens.
+ * Examples: "org.freedesktop.policykit.exec", "org.freedesktop.PolicyKit1.test"
+ *
+ * Returns: TRUE if @action_id is valid, FALSE otherwise.
+ */
+static gboolean
+_validate_action_id (const gchar *action_id)
+{
+  gsize n;
+  gsize len;
+  gboolean at_component_start;
+
+  if (action_id == NULL)
+    return FALSE;
+
+  len = strlen (action_id);
+  if (len == 0)
+    return FALSE;
+
+  /* Must contain at least one dot (reverse-DNS requires >= 2 components) */
+  if (strchr (action_id, '.') == NULL)
+    return FALSE;
+
+  at_component_start = TRUE;
+  for (n = 0; n < len; n++)
+    {
+      gchar c = action_id[n];
+
+      if (c == '.')
+        {
+          /* No leading, trailing, or consecutive dots */
+          if (n == 0 || n == len - 1 || action_id[n - 1] == '.')
+            return FALSE;
+          at_component_start = TRUE;
+          continue;
+        }
+
+      if (at_component_start)
+        {
+          /* Each component must start with an ASCII letter (upper or lower) */
+          if (!((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z')))
+            return FALSE;
+          at_component_start = FALSE;
+          continue;
+        }
+
+      /* Subsequent chars: ASCII letter, digit, or hyphen */
+      if (!((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '-'))
+        return FALSE;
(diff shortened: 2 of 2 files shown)


jrybar-rh:
## Code Review: PR #676 — security: fix three hardening issues found by static analysis

**Files reviewed:** `src/polkit/polkitunixprocess.c`, `src/polkitagent/polkitagenthelperprivate.c`, `src/polkitbackend/polkitbackendactionpool.c`

Three independent bug-fix commits are bundled here. Fixes 1 and 2 are broadly sound; Fix 3 has two actionable issues detailed below.

---

### Critical

*(none)*

---

### High

**[3a] `_validate_action_id()` breaks the existing integration test — `polkitbackendactionpool.c` (new function)**

The integration test at `test/integration/pkexec/test.sh` (lines 166, 208) registers:

```xml
<action id="org.freedesktop.PolicyKit1.test.run">
```

The component `PolicyKit1` starts with an uppercase letter (`P`). `_validate_action_id()` rejects any component whose first character is not `[a-z]`, so this action ID fails validation, `_start()` calls `goto error`, and `XML_StopParser()` aborts the parse. The integration test would fail silently or with a confusing warning.

More broadly, third-party `.policy` files in the wild (udisks2, NetworkManager, etc.) have historically included uppercase components in their action IDs. Activating this stricter validation without a documented deprecation period risks rejecting valid installed policy files on upgrade.

**Suggested fix:** Either update the integration test to use a lowercase-only action ID (`org.freedesktop.polkit.test.run`), or relax the validator to allow `[A-Za-z]` at component start (matching the historical convention), or gate the validator on a new strict-mode option.

---

**[3b] Missing forward declaration causes a C99 compile error — `polkitbackendactionpool.c`**

`_start()` is defined at line ~740–887. The new `_validate_action_id()` is placed *after* `_start()` (inserted between lines 887 and 889). In C99 (which this project builds with, per `meson.build` line 8: `c_std=c99`), calling an undeclared function is an error, not a warning. The build already enables `-Wimplicit-function-declaration` (meson.build line 108).

Without a forward declaration added before `_start()`, `make` / `ninja` will fail to compile `polkitbackendactionpool.c`.

**Fix:** Add before `_start()`:
```c
static gboolean _validate_action_id (const gchar *action_id);
```
or move the `_validate_action_id()` definition *before* `_start()`.

---

### Medium

**[2b-incomplete] Other `error->message` NULL-deref patterns in `polkitagenthelperprivate.c` left unpatched**

The PR fixes the `send_dbus_message()` failure path at line 131 (post-patch), but the same GError pattern exists at lines 100 and 108 (both unmodified):

```c
// line 100 — polkit_authority_get_sync failure
g_printerr ("Error getting authority: %s\n", error->message);
g_error_free (error);

// line 108 — polkit_unix_user_new_for_name failure
g_printerr ("Error constructing identity: %s\n", error->message);
g_error_free (error);
```

`polkit_authority_get_sync()` can return `NULL` from `get_uninitialized_authority()` in a code path that does not call any error-setting function (the singleton allocation path). If `error` is left `NULL` on those callers, dereferencing `error->message` is a NULL pointer dereference. The PR should either guard all three sites consistently or note explicitly why lines 100/108 are safe.

---

### Low

**[1] `polkit_unix_process_set_pid()` bug fix — correct and complete**

Passing `pid` (the parameter) instead of `process->pid` (stale) to `SYS_pidfd_open` is clearly correct. The PR description's analysis of the masking condition (object freshly constructed → `process->pid == 0` → `pidfd_open(0)` returns `ESRCH` → fallback) is accurate. No issues here.

**[2a] `read_cookie()` OOB guard — correct**

Storing `strlen(buf)` in `len` and checking `len == 0` before indexing is the right fix. The `size_t` declaration mid-block is valid C99. The drain-and-reject path for `len == 0` is appropriate conservative behavior. No issues here.

---

### Nitpick

**[3c] `_validate_action_id()` uses `gsize n` / `gsize len` instead of `guint n` / `gsize len`**

`_validate_icon_name()` immediately below uses `guint n` as the loop variable. For style consistency with the surrounding file, consider using `guint` (or `gsize`) consistently. This is a minor style note only.

**[3d] The error path in `_start()` only calls `XML_StopParser()` — no `g_warning()` on the action ID itself**

The warning is emitted by `_validate_action_id()`'s caller (`g_warning ("Action id '%s' is invalid", attr[1])`), which is correct, but the comment in `_start()`'s error label makes no mention of what triggered the abort. This is fine as-is but could be slightly confusing during debugging.

---

## Summary

| # | Severity | Issue |
|---|----------|-------|
| 3a | **High** | Validator rejects `org.freedesktop.PolicyKit1.*` — breaks integration test and potentially production policy files |
| 3b | **High** | Missing forward declaration for `_validate_action_id()` → C99 compile error |
| 2b | Medium | `error->message` NULL-deref at lines 100/108 left unfixed while line 131 is fixed |
| 1 | ✓ Low (fixed) | `pidfd_open` wrong variable — fix is correct |
| 2a | ✓ Low (fixed) | `buf[-1]` OOB guard — fix is correct |
| 2b | ✓ Low (fixed) | `send_dbus_message` NULL deref at line 131 — fix is correct |

---

rawrmonster17:
Thanks for the review. All points addressed in the updated commit:

**High [3a] — validator rejects `org.freedesktop.PolicyKit1.*`:**
`_validate_action_id` now accepts `[A-Za-z]` at component start and `[A-Za-z0-9-]` in subsequent positions, matching historical convention and the D-Bus interface name format polkit action IDs are modelled on. `org.freedesktop.PolicyKit1.test.run` passes. The doc comment is updated to reflect this.

**High [3b] — missing forward declaration / implicit declaration:**
`_validate_action_id` is now defined before `_start` in the file (inserted at line 737, `_start` begins at ~800). No implicit declaration; the definition precedes all uses.

**Medium [2b] — error->message at lines 100 and 108 left unguarded:**
Both sites are now guarded with `error != NULL ? error->message : "(no error message)"` and use `g_clear_error` instead of `g_error_free`. All three `error->message` dereferences in `polkitagenthelperprivate.c` (lines 100, 108, 131) are now consistently handled.

PR #692, opened 2026-08-08
polkitd: add support for systemd watchdog

### LINKED ISSUE #690: WatchdogSec= support for polkitd
**Is your feature request related to a problem? Please describe.**
polkitd can be in the critical path of a systemd host gating privileged operations using polkit rules, in the event polkitd gets hung up for any reason - it can result in a substantially degraded machine.

**Describe the solution you'd like**
Support for WatchdogSec= / sd_notify("WATCHDOG=1") in polkitd to forcibly restart the daemon if hung 

Implementing it myself in our private tree, trying the ask upstream approach first.

**Additional context**
We've started using polkit more and are seeing what seems like occasional deadlocks involving the dbus/polkit machinery for polkit-governed `systemctl` operations.

The lack of service watchdogs in both polkitd and dbus-daemon (not part of polkit, I know), seems like a gaping hole in the face of these types of failures.

Is there a particular reason these daemons haven't received the service watchdog treatment upstream already?

bluca:
WatchdogSec= can be set in the unit, and if so the service has to send a periodic ping or it will be terminated. This provides guarantees against the service getting stuck.

Do not configure it for now, as it is unclear what an acceptable high ceiling could be due to custom user-supplied rules. Users can experiment with the feature by simply adding a local drop-in to enable it.

Fixes https://github.com/polkit-org/polkit/issues/690

### DIFF
--- a/src/polkitbackend/polkitd.c
+++ b/src/polkitbackend/polkitd.c
@@ -201,6 +201,15 @@ on_sigint_sigterm (gpointer user_data)
   return TRUE;
 }
 
+#ifdef HAVE_LIBSYSTEMD
+static gboolean
+on_watchdog_timeout (gpointer user_data)
+{
+  sd_notify (0, "WATCHDOG=1");
+  return G_SOURCE_CONTINUE;
+}
+#endif
+
 static gboolean
 on_sighup (gpointer user_data)
 {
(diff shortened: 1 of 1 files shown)


vcaputo-eero (approved):
lgtm, thanks @bluca

PR #694, opened 2026-08-12
polkitagent: Fix use-after-free in io_watch_have_data()

jrybar-rh:
complete_session() emits the "completed" signal, and handlers of that signal are explicitly documented (and, in the case of PolkitAgentTextListener, actually implemented) to be allowed to drop the last reference to the PolkitAgentSession.

io_watch_have_data() can call complete_session() once from one of its error/success branches and then unconditionally call it a second time from its epilogue whenever the dispatch also carries G_IO_ERR or G_IO_HUP. If the first call already caused the session to be freed via its "completed" handler, that second call operates on freed memory.

Fix this by NULLing out the local session pointer right after each complete_session() call, and skipping the epilogue's call when it is already NULL, so we never dereference a session that may have already been freed.

Comments and description co-authored-by @claude 
Reported by: aisle.com

### DIFF
--- a/src/polkitagent/polkitagentsession.c
+++ b/src/polkitagent/polkitagentsession.c
@@ -444,6 +444,13 @@ io_watch_have_data (GIOChannel    *channel,
       g_warning ("in io_watch_have_data() but helper is not supposed to be running");
 
       complete_session (session, FALSE);
+      /* complete_session() emits ::completed, and handlers of that signal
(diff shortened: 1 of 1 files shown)


vmihalko (approved):
The NULL-out pattern is the standard GLib idiom for this. All paths are covered. LGTM!

PR #703, opened 2026-09-04
ci: switch the Alpine job to the edge branch

mrc0mmand:
Apart from catching potential issues earlier, this should also finally fix the outstanding build fail in our Alpine CI job, as we need newer fortify-headers package with [0] that's currently only in the edge branch.

[0] https://github.com/jvoisin/fortify-headers/commit/b9121e4d679b5fccb69abcb58b05a6be5c7191dc

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
   alpine:
     runs-on: ubuntu-latest
     container:
-      image: alpine:latest
+      image: alpine:edge
       options: "--privileged"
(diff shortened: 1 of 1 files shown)


mrc0mmand:
The remaining Testing Farm errors are, unfortunately, outside of our reach as they're caused by a bug in dnf on Fedora Rawhide/45/44, see https://status.testing-farm.io/issues/2026-09-02-failing-installations-from-copr-repositories-on-fedora/. So we'll have to wait that one out.

